PR #64, opened 2022-12-08
Cleanup Tests: #bytecodes check Pharo10 vs Pharo11 not needed

MarcusDenker:
I backported the #bytecodes methot to Pharo10, so we can remove the versions checks


MarcusDenker:
Will redo as part of https://github.com/ApptiveGrid/Soil/pull/65 which touches the same tests

PR #73, opened 2022-12-09
Some timezones are 30 or 45 minutes off the hour

yannij:
Change the Date serialize/materialize to save offset in minutes, instead of hours. Add a test case for offset that is not a whole hour. Also test that a Date serialized in one timezone, can be materialized in a different local timezone, and is still equal.

### DIFF
--- a/src/Soil-Core-Tests/SoilPrimitiveSerializationTest.class.st
+++ b/src/Soil-Core-Tests/SoilPrimitiveSerializationTest.class.st
@@ -12,6 +12,14 @@ SoilPrimitiveSerializationTest >> newSerializer [
 	^ Soil inMemory newSerializer
 ]
 
+{ #category : #helpers }
+SoilPrimitiveSerializationTest >> restoreLocalTimeZoneAfter: aBlock [
+
+	| realTimeZone |
+	realTimeZone := DateAndTime localTimeZone.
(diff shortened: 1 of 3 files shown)


yannij:
Forgot to fix other testcase affected by the previous change. Also, have not done pull requests before, so it may be messier than it needed to be.

noha:
Thanks for the pull request. Is there any reason why use the detour over system TimeZone to create a Date with a special offset?

yannij:
DateTest and DateAndTimeTest use #useTimeZone:during:, but the abbreviation list is incomplete and only has whole hour offsets. See TimeZone class>>#timeZones. In testSerializationDateWithLocalTimeZone, a similar #useTimeZoneInstance:during: is used, so the TimeZone's with non-whole hour offsets could be tested.

To use Date class>>starting: you end up in a similar place because the DateAndTime argument that is needed, also ends up using the local time zone. I did it this way: (1) it's similar to how DateTest and DateAndTimeTest do it; (2) I was skeptical that code such as "(Date fromDays: 10) offset: -5 hours" would be equivalent to the normal way I'd create with Date class>>#year:month:day:

noha:
I will need to make a few adjustments but will merge now. No need to wait

PR #89, opened 2022-12-20
[WIP] Support behavior versions

noha:


src/Soil-Core/SoilMaterializer.class.st
```diff
@@ -37,16 +37,28 @@ SoilMaterializer >> materializeFromBytes: aByteArray [
...
+	^ description isCurrent 
+		ifTrue: [ 
+			objectClass classLayout soilBasicMaterialize: objectClass with: self ]
+		ifFalse: [ 
+			versions := transaction behaviorVersionsUpTo: description.
+			obj := objectClass basicNew.
+			self registerObject: obj.
+			versions last instVarNames do: [ :instVar | | value |
+				value := self nextSoilObject.
+				(versions allSatisfy: [ :version | version instVarNames includes: instVar ]) ifTrue: [
```
MarcusDenker, line 58: Do I understand this correctly?
the allSatisfy: check is needed as if the var would be removed at some point int the past, it should be nil, even though there is a new var with the same name.

MarcusDenker (approved):

PR #98, opened 2022-12-22
add testMaterializingObjectWithReOrderedIvars

MarcusDenker:
this test checks that we can read objects where the order of the ivars changed. This happens for example when moving ivars up or down in the hierarchy


MarcusDenker:
Not good... lots of uncategorized moved wrongly to "accessing".

I will redo it

PR #99, opened 2022-12-22
automatic whitespace cleanup

MarcusDenker:
run the script to cleanup whitespace


MarcusDenker:
Ah, so this moves all uncategories to "accessing". Bad. Closed

PR #122, opened 2023-01-07
Fix: committing objects whose class have changed

MarcusDenker:
This PR makes sure that we can commit after a class changed shape, see #testMaterializingObjectWithIvarRemovedThenCommit

What was missing? We get the behavior description via #behaviorDescriptionFor: from the behaviorRegistry, but that one is the old one (with the ivar not removed) Solution: create a new SOBehaviorDescription with the same objectID. 

This has to be moved to the registry later so we can find the history of a desciption, but it makes committing objects with changed classes possible now The code in behaviorDescriptionFor: will be refactored later.

Small changes not related:

- testSOBehaviorDescription
- use SOBehaviorDescription meta", add #metaId that returns the id with index 2

### DIFF
--- a/src/Soil-Core/SOTransaction.class.st
+++ b/src/Soil-Core/SOTransaction.class.st
@@ -45,32 +45,34 @@ SOTransaction >> atObjectId: objectId putObject: anObject [
 { #category : #public }
 SOTransaction >> behaviorDescriptionFor: aClass [
 	| classDescription id |
-	classDescriptions 
-		at: aClass soilBehaviorIdentifier 
+	classDescriptions
+		at: aClass soilBehaviorIdentifier
 		ifPresent: [ :description | ^ description ].
-	soil behaviorRegistry 
-		nameAt: aClass soilBehaviorIdentifier 
+	soil behaviorRegistry
+		nameAt: aClass soilBehaviorIdentifier
 		ifPresent: [ :oid | | desc |
-			desc := (oid index = 2) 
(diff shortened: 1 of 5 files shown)


src/Soil-Core/SOTransaction.class.st
```diff
@@ -45,32 +45,34 @@ SOTransaction >> atObjectId: objectId putObject: anObject [
...
 		ifPresent: [ :oid | | desc |
-			desc := (oid index = 2) 
-				ifTrue: [(SOBehaviorDescription for: SOBehaviorDescription) objectId: oid ]
-				ifFalse: [ (self objectWithId: oid ifNone: [Error signal ]) objectId: oid; yourself ].
+			desc := (oid index = 2)
+				ifTrue: [SOBehaviorDescription meta objectId: oid ]
+				ifFalse: [ | foundDesc |
+					 "the description in the database might not be current, if not, we create a new one with the same ID"
+					 foundDesc := (self objectWithId: oid ifNone: [Error signal ]) objectId: oid; yourself.
+					 foundDesc isCurrent ifFalse: [ (SOBehaviorDescription for: aClass) objectId: oid ] ifTrue: [foundDesc]].
```
noha, line 59: Is it necessary to put an objectId: there? The current concept includes having different objectIds for different behavior versions. So this looks a bit cheesy

MarcusDenker: yes, this is odd... I will look.

PR #136, opened 2023-01-14
Use only zero index

noha:
Serialize with segment zero index which is preknown. Don't store behaviors as external reference, this leads just to a lot of data to be stored


noha:
@MarcusDenker The failing test indicates a problem with the handling of isCurrent when storing an object. Assigning the same objectId for a new behavior is wrong. Also missing is that on write a not current behaviorDescription needs to made added to the behaviorRegistry

noha:
should be merged after #135

MarcusDenker (approved):
I fixed it in the spirit of the PR: the description is now created with a new ID. This way we can merge this PR before changing to use the same ID + class version

PR #169, opened 2023-01-28
[WIP] Fix fnctl locking

noha:
I just discovered that lock calls go with from:to: but in the end the fcntl call uses from:length: and the to: argument is just passed as length


noha:
This still needs to be checked on linux about locks have proper offsets. On MacOS there seems to be no way to acquire information about file locks in the system

PR #177, opened 2023-01-31
Contributions

astares:
Small cleanups

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# Bash needs lf even on windows/cygwin
+*.st eol=lf
+*.sh eol=lf
+
+# st files are always text, even if they contain invalid encoded chars
+*.st text diff
+
+# GitHub is unable to properly detect Smalltalk code, so help it a bit
+*.st linguist-language=Smalltalk
--- a/.smalltalkci/.unit-tests.ston
+++ b/.smalltalkci/.unit-tests.ston
@@ -1,0 +1,16 @@
+SmalltalkCISpec {
+  #loading : [
+    SCIMetacelloLoadSpec {
+      #baseline : 'Soil',
+      #directory : '../src',
+      #load : [ 'CI' ],
+      #platforms : [  #pharo ]
+    }
+  ],
+  #testing : {
+    #coverage : {
+      #packages : [ 'Soil*' ],
+      #format: #lcov
+    }
+  }
+}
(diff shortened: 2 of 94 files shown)


noha:
The signal-to-noise ratio is extremely low in this PR. The main contribution as I can see is the git attributes. This PR gets only merged if it does not conflict with any local copy we have right now. Please make smaller PRs and leave out the whitespace and dot cleanup. We don't need that

astares:
PR's for each small issue would be a lot of work  

Beside the .gitattributes it includes:
 - typo fixing 
 - fixing the example in SOObjectId class >> example by switching from <example> to <sampleInstance> pragma so it is clickable again
 - proper categorization
 - lint issues like banning the "Access to system" class
 - proper use of groups in Baselines (uppercase like "Test" and "Core" for own, using lowercase "default" and "all" for the predefined virtual groups of Metacello)
 - having a "CI" group in the baseline  
 - now also the Smalltalk CI files needed for the coverage (which I forgot to commit yesterday)
 
and it's up to you what you might see as useful. The whitespaces make no difference, but if you need you dots and do not like such cleanings then ignore the PR.  

Nonetheless you might want to grab a few of the contributions manually if you see them as useful. Unfortunately Iceberg does not allow for cherry-picking...

noha:
Small commits and small PRs are the best assurance to get things integrated. If it is too much work I can translate that to "it is not worth it". You put the work now on my side which is why will shortly cherry pick the stuff that comes to my eyes and discard the rest. It is hard to do a database beside my usual work and if you put additional work on my shoulder I'm not gonna like it

astares:
Hi Norbert 

I understand and you are right. Sorry for not being able to provide it as small PRs yesterday, at least I tried to be descriptive in text of each commit. Will try to do it better next time. Unfortunately time is always the limiting factor. 😔  
(I sideways want to write a small tool to make my life easier and wanted to check if Soil persistency could help me to persist. I though easiest would be to browse through and learn about it from the code and tests and while doing that I noticed a few things leading to the PR).

noha:
Don't worry. I'm happy you try out soil and give input. But it is still in ramping up phase and it is not easy to do so hacks and unclean code are obviously the result of it. So cleaning while being always something good is not good this time when it distracts me. So I'm dependent on small PRs with clear goal. If I have to dive in to understand what the PR does in order to see if it should be integrated it distracts me and so in this case I rather not integrate. 
About the whitespace and dot cleaning I don't see it as an improvement. What is the gain? Uniformity of code characters is not an asset in my opinion. So here and in pharo it does something valuable but at the same time there are tons of PRs for that using the CI pipeline and someone has to integrate and merge. So they drawbacks are clear the benefit not so much. So I hope you can see that this can be objected to.

PR #181, opened 2023-02-03
 102-Migration-support-class-rename 

MarcusDenker:
A first very simple class rename support

If a class is renamed, we have to tell soil by calling #renameClassNamed: to:

This can be called as part of a migration script, for example.

### DIFF
--- a/src/Soil-Core/Soil.class.st
+++ b/src/Soil-Core/Soil.class.st
@@ -140,3 +140,28 @@ Soil >> path [
 Soil >> path: aString [ 
 	path := aString asFileReference
 ]
+
+{ #category : #refactorings }
+Soil >> renameClassNamed: oldName to: newName [
+	| transaction behaviorDescription objectId |
+
+	"we load the behaviorDescription of the oldName, change the name and commit"
+
+	transaction := self newTransaction.
+
+	objectId := self behaviorRegistry
+		nameAt: oldName
+		ifAbsent: [ self error: 'name not found in behavior registry' ].
+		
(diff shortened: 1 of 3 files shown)


MarcusDenker:
This does not increase the version of the BehvaviorDescription, I think this should be done but requires some changes to 
behaviorWithIndex:andVersionifNone: before

MarcusDenker:
To be checked: we should update the behavior registry to have an entry for the new name

src/Soil-Core/Soil.class.st
```diff
@@ -140,3 +140,28 @@ Soil >> path [
...
+
+{ #category : #refactorings }
+Soil >> renameClassNamed: oldName to: newName [
+	| transaction behaviorDescription objectId |
+
+	"we load the behaviorDescription of the oldName, change the name and commit"
+
+	transaction := self newTransaction.
+
+	objectId := self behaviorRegistry
```
noha, line 152: Regarding version bump: The mechanic is so that the name always points to the newest version of a behavior description. We decided to keep the object id for a behavior description and introduce versions. So taking that at this point it is the newest and a simple call to #incrementVersion should be sufficient to have the version update

noha:
> To be checked: we should update the behavior registry to have an entry for the new name

Yes, I think we need to add the generated behaviorDescription in the name index with the new name. And we might remove the entry for the old name because it isn't used anymore. Reading uses the object id to resolve the behavior description and if we always migrate on read then the old name is never used

src/Soil-Core/Soil.class.st
MarcusDenker: done

noha (approved):

PR #184, opened 2023-02-11
unify how the objects ivar is initialized in the SoilObjectCodec Hierarchy

MarcusDenker:
This PR unified how the objects ivar is initialized in the SoilObjectCodec hierarchy:

 it was lazy in one and at creation for the other. We can set it on creation ones for both.

- move #initialize up to SoilObjectCodec
- remvove nil check in #registerObject: (which is called for every object, so it should be fast)

### DIFF
--- a/src/Soil-Serializer/SoilMaterializer.class.st
+++ b/src/Soil-Serializer/SoilMaterializer.class.st
@@ -215,7 +215,6 @@ SoilMaterializer >> nextSoilObject [
 
 { #category : #registry }
 SoilMaterializer >> registerObject: anObject [
-	objects ifNil: [ objects := OrderedCollection new ].
 	objects add: anObject 
 ]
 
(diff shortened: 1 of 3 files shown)


src/Soil-Serializer/SoilObjectTable.class.st
```diff
@@ -0,0 +1,32 @@
...
+
+{ #category : #initialization }
+SoilObjectTable >> initialize [
+	lastIndex := 0.
+	index := IdentityDictionary new
+]
+
+{ #category : #accessing }
+SoilObjectTable >> size [
+	^ lastIndex
```
noha, line 31: Is this much faster than "index size"?

MarcusDenker: ah, good idea...the size here is just used in tests, but using the size of of the dictioary instead of managing lastIndex would be nice. (in the add: method)

PR #189, opened 2023-02-15
146-Refactor-behavior-registry-to-load-and-hold-all-versions-of-a-behavior 

MarcusDenker:
use the "versions" ivar of the behaviorRegistry to cache versions index -> chain of behavior history

- unify behaviorVersionsUpTo:  and behaviorDescriptionWithIndex: andVersion: to first fill the cache and use it
- remove #isZero check in #nextBehaviorDescription

### DIFF
--- a/src/Soil-Core/SoilBehaviorRegistry.class.st
+++ b/src/Soil-Core/SoilBehaviorRegistry.class.st
@@ -20,7 +20,7 @@ SoilBehaviorRegistry >> addSpecialObjects [
 SoilBehaviorRegistry >> behaviorDescriptionWithIndex: behaviorIndex andVersion: version transaction: transaction [
 	| versionsOfBehavior |
 
-	self loadHistoryForBehaviorWithId: (SOObjectId segment: 0 index: behaviorIndex) transaction: transaction.
+	self loadHistoryForBehaviorWithIndex: behaviorIndex transaction: transaction.
 
 	versionsOfBehavior := versions at: behaviorIndex ifAbsent: [self halt].
 
@@ -35,7 +35,7 @@ SoilBehaviorRegistry >> behaviorVersionsUpTo: aSOBehaviorDescription transaction
 	objectId := self 
 		nameAt: aSOBehaviorDescription behaviorIdentifier 
 		ifAbsent: [ self halt ].
-	self loadHistoryForBehaviorWithId: objectId transaction: transaction.
+	self loadHistoryForBehaviorWithIndex: objectId index transaction: transaction.
 	chain := versions at: objectId index.
 	chain first isCurrent ifFalse: [ 
 		chain addFirst: ((SOBehaviorDescription for: aSOBehaviorDescription objectClass) version: (aSOBehaviorDescription version + 1))].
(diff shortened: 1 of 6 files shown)


src/Soil-Core/SoilBehaviorRegistry.class.st
```diff
@@ -16,24 +16,29 @@ SoilBehaviorRegistry >> addSpecialObjects [
 		put: SOBehaviorDescription metaId
 ]
 
+{ #category : #'as yet unclassified' }
+SoilBehaviorRegistry >> behaviorDescriptionWithIndex: behaviorIndex andVersion: version transaction: transaction [
+	| versionsOfBehavior |
+
+	self loadHistoryForBehaviorWithId: (SOObjectId segment: 0 index: behaviorIndex) transaction: transaction.
```
noha, line 23: metaAt: had the purpose that if the segement is known there is no need to create an object id all the time. We need to make a distinction about **id** being the objecId (maybe call it always like that). The other is the **index** which is the index within a segment if the segment is known which is always the case for segment zero

src/Soil-Core/SoilBehaviorRegistry.class.st
```diff
@@ -16,24 +16,29 @@ SoilBehaviorRegistry >> addSpecialObjects [
...
-			materializeObject;
-			object  ].
+	self loadHistoryForBehaviorWithId: objectId transaction: transaction.
+	chain := versions at: objectId index.
 	chain first isCurrent ifFalse: [ 
-		chain addFirst: (SOBehaviorDescription for: aSOBehaviorDescription objectClass) ].
-	versions 
-		at: aSOBehaviorDescription behaviorIdentifier 
-		put: chain.
+		chain addFirst: ((SOBehaviorDescription for: aSOBehaviorDescription objectClass) version: (aSOBehaviorDescription version + 1))].
```
noha, line 41: What happens in case the code is newer than the one in the database? On read time we add the current version on top but on writing an object with the new version who is storing the behavior description. Can it be that we end up with an object in database that has a behavior description version missing?

src/Soil-Core/SoilBehaviorRegistry.class.st
MarcusDenker: I changed it to not create an object id to read the behavior historu

src/Soil-Core/SoilBehaviorRegistry.class.st
MarcusDenker: this is only done if the code is newer (if it changed):

```
	chain first isCurrent ifFalse: [ 
		chain addFirst: ((SOBehaviorDescription for: aSOBehaviorDescription objectClass) version: (aSOBehaviorDescription version + 1))].
```

noha: That is what I meant. I you first read an object with a newer code the new behavior description is added to the top of the chain. But if after that an object is written the behavior description does not get written to database, right?

MarcusDenker: I improved SoilMigrationTest>>#testMaterializingObjectWithIvarRemovedThenCommit, we store an object, then remove an ivar, then load it back, then we commit it again and it works:

```
testMaterializingObjectWithIvarRemovedThenCommit
	"We can load an object that was saved with two ivars even if the current class has just one"
	| tx tx2 materializedRoot object |
	object := migrationClass new.
	object
		instVarNamed: #one put: 1;
		instVarNamed: #two put: 2.
	tx := soil newTransaction.
	tx root: object.
	tx commit.
	migrationClass
		removeSlot: (migrationClass slotNamed: #two).
	
	"load it"	
	tx2 := soil newTransaction.
	materializedRoot := tx2 root.
	self assert: materializedRoot class instVarNames size equals: 1.
	self assert: (materializedRoot instVarNamed: #one) equals: 1.

	"lets try to commit with the class changed"
	tx := soil newTransaction.
	tx root: materializedRoot.
	tx commit.

	"load again"
	tx2 := soil newTransaction.
	materializedRoot := tx2 root.
	self assert: materializedRoot class instVarNames size equals: 1.
	self assert: (materializedRoot instVarNamed: #one) equals: 1
```

noha: Sure it works because the change class is still the same as when writing it. So a missing behavior description in the database cannot be seen because the one created from memory will be the same. The test should test if there are two versions of that behavior are in the database

PR #205, opened 2023-02-22
192-Should-SOObjectProxy-inherit-from-ProtoObject

### LINKED ISSUE #192: Should SOObjectProxy inherit from ProtoObject?
if we inherit from ProtoObject, we will forward more methods transparently

MarcusDenker:
- Change superclass of Proxy to ProtoObject
- Fix tests as the proxy now is much better 

fixes #192

### DIFF
--- a/src/Soil-Core-Tests/SoilPersistentDictionaryTest.class.st
+++ b/src/Soil-Core-Tests/SoilPersistentDictionaryTest.class.st
@@ -56,8 +56,6 @@ SoilPersistentDictionaryTest >> testCommitAndRead [
 
 	| dict transaction materialized |
 	
-	"to remove"
-	SoilTypeCodes initialize.
 	transaction := soil newTransaction.
 	dict := SoilPersistentDictionary new soilMakePersistentIn: transaction.
 	dict add: (#test -> 'string').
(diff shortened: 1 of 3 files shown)


src/Soil-Core-Tests/SoilTest.class.st
```diff
@@ -97,7 +97,7 @@ SoilTest >> testSerializingRootTwoCluster [
 	tx checkpoint.
 	tx2 := soil newTransaction.
 	materializedRoot := tx2 root.
-	self assert: materializedRoot nested reference class equals: SOObjectProxy.
+	self assert: materializedRoot nested reference class equals: SOTestClusterRoot.
```
noha, line 100: This is wrong. The nested reference returns a proxy and even ProtoObject can answer class

src/Soil-Core-Tests/SoilPersistentDictionaryTest.class.st
```diff
@@ -66,10 +64,7 @@ SoilPersistentDictionaryTest >> testCommitAndRead [
 
 	transaction := soil newTransaction.
 	materialized := transaction root.
-
-	"should this not be transparently load it?"
-	"if the Proxy inherits from ProtoObject, it works"
-	self assert: (materialized at: #test) class equals: SOObjectProxy
+	self assert: (materialized at: #test) equals: 'string'
```
noha, line 67: I don't know why the objects are being resolved but it is wrong. This defeats the purpose of having proxies

MarcusDenker:
So it is the send of #yourself in SOClusterRecord>>#referenceAt:

I will do a new PR

PR #207, opened 2023-02-22
Unify encoding/decoding WideSymbols/WideStrings

MarcusDenker:
unify how WideString and WideSymbol are encoded  (use the encoder defined for both)


src/Soil-Serializer/SoilBasicMaterializer.class.st
```diff
@@ -145,8 +145,7 @@ SoilBasicMaterializer >> nextTime [
 { #category : #reading }
 SoilBasicMaterializer >> nextWideString [
 	| buf length wideString |
-	length := self nextLengthEncodedInteger.
-	buf := ByteArray new: length.
+	buf := ByteArray new: (length := self nextLengthEncodedInteger).
```
noha, line 148: I don't find that this improves readability

noha (approved):

PR #208, opened 2023-02-23
193-SOTransaction-idMap-use-atifPresent 

### LINKED ISSUE #193: SOTransaction idMap: use #at:ifPresent:
The idMap is a dictionary objectID -> record, but we search it using #detect:ifFound:, which does a linear scan of the records:

```
objectWithId: objectId ifNone: aBlock
	idMap 
		detect: [ :each | each objectId = objectId ]
		ifFound: [ :record | ^ record object ].

	^ self materializationWithId: objectId ifNone: aBlock.
```

We can use instead at:ifPresent:, to use the dictionary key and avoid linear scan:

```
objectWithId: objectId ifNone: aBlock
	idMap 
		at: objectId 
		ifPresent: [ :record | ^ record object ].

	^ self materializationWithId: objectId ifNone: aBlock.
```

MarcusDenker:
We can query the idMap by identity

- The behaviorDescriptionWithIndex: does not need to read this cache as we cache the versions in the behaviorRegistry
- #objectWithId: can  use #at:ifPresent: to avoid lineary scan

remove two "self halt" that are not needed anymore

fixes #193

### DIFF
--- a/src/Soil-Core/SOTransaction.class.st
+++ b/src/Soil-Core/SOTransaction.class.st
@@ -112,13 +112,7 @@ SOTransaction >> behaviorDescriptionFor: aClass [
 ]
 
 { #category : #'as yet unclassified' }
-SOTransaction >> behaviorDescriptionWithIndex: index andVersion: version [ 
-	
-	idMap 
-		detect: [ :each | each objectId = index ]
-		ifFound: [ :record | 
-			(record object version = version)
-				ifTrue: [ ^ record object ] ].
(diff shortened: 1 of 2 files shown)


noha:
We need to handle the behaviors in the transaction as we do not want to change the behavior registry before committing. That will introduce subtle side effects if one transaction adds something in the registry another transaction relies on that information and then the first transaction is aborted. All these scenarios could be prevented if we keep new behaviors in the transaction beforre commiting

MarcusDenker:
I close this

PR #244, opened 2023-04-04
 Indexing infrastructure #4

noha:
Refactoring the storage of pages so it performs well. Clean the interface


noha:
This is the next step for indexing. Copy on write is now reifified in the model and can be used efficiently. Setup of skip list dictionary is still not perfect but works for now.

MarcusDenker (approved):

PR #251, opened 2023-04-08
 Fix foreign tests #3

noha:
Next round fixing foreign tests


noha:
@MarcusDenker Another PR I will merge because it became too big and blocks progress

PR #252, opened 2023-04-08
Add initial version of Soil-PlantUml package

yannij:
The code done with method extensions, so deployment can exclude this diagram generation code. Maybe it isn't worth the bother of an extra package though.

### DIFF
--- a/src/Soil-PlantUml/Object.extension.st
+++ b/src/Soil-PlantUml/Object.extension.st
@@ -1,0 +1,26 @@
+Extension { #name : #Object }
+
+{ #category : #'*Soil-PlantUml' }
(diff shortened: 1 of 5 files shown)


noha:
Thanks! This needs quite some effort to establish the diagram. For now I don't want to include this in the main repo. Everything in here needs to be actively maintained. So I will merge that an a separate branch for later. If documentation comes into focus I will have another look and try to make more things discovered automatically.

yannij:
Sounds right. Documentation should explain the design, but code is the implementation. Trying to derive the design from the code is probably always going to yield something that won't be good explanatory docs. Anyhow, when code and design is changing rapidly, having the docs try to track the changes probably won't work out.

Issue #121, opened 2023-01-05
Add Btree+ index


noha:

MarcusDenker:
TODO:

- [ ] Test:  Index Split when rootpage is full
- [ ] Test: index Split of non-root page
- [ ] Storage is incomplete (misses items write, tests)
- [ ] For loading we need to add pageCode support to be able to load index pages
- [ ] Test: write btree with index+multiple data pages
- [ ] removal not yet supported

DONE:

- [ ]  we can add to the data page
- [ ] data page split adds index entry
- [ ] index is used for at: and at:put:
- [ ] #size: implemented in iterator and test

noha:
Those todo entries should be issues

PR #266, opened 2023-04-21
B+Tree 1

MarcusDenker:
- First B+Tree, incomplete:
	- Two kinds of pages: Data and Index
	- headerPage is the first Data page
	- rootPage is the root index page
	- Data Pages are  linked list for fast in-order iteration
	- Index stores key -> PageID
	
- What works
	- we can add to the data page
	- data page split adds index entry 
	- index is used for at: and at:put:
	
TODO
	- Index Split when full, especially check rootPage
	- removal not yet supported


MarcusDenker:
I added the TODO to https://github.com/ApptiveGrid/Soil/issues/121

noha (approved):

PR #363, opened 2023-06-13
Initialize database and object counter with zero so the first increme…

noha:
…nt gives 1


noha:
Just wondering why it fails now

MarcusDenker:
It fails as this was merged today: https://github.com/pharo-project/pharo/pull/13965

noha:
Ok, then I will add another or: case so we do not block further development

MarcusDenker (approved):
ok

PR #367, opened 2023-06-15
SoilIndexPage: move ivars and methods down

MarcusDenker:
- move ivars "items keySize" from SoilSkipListPage and SoilBTreePage to the superclass SoilIndexPage
- this allows us to move lots of duplicated methods there, too
- add some #subclassResponsibility methods


noha:
This is most probably not a good move. An index page is not only for keeping items. An index page that manages free lists of pages is different

PR #379, opened 2023-06-21
this fixes inspecting the journal from a soil instance.

MarcusDenker:
SoilJournal>>#inspectionContent was using #allTransactionJournals, but there is only #transactionJournals

in addition, categorize the inspector methods as 'tools'


MarcusDenker:
ups, already done as https://github.com/ApptiveGrid/Soil/pull/378

PR #383, opened 2023-06-22
make journal partial replayable. Changing management of journals not …

noha:
…to read from disk all the time.

Now it is cached and bogus. Next steps need to elaborate invalidating the cache


MarcusDenker (commented):
ok

PR #391, opened 2023-06-26
- nextAfter kept stopping once getting to the last item in a page.

noha:
- keyOrClosestAfter treated anindex as element


MarcusDenker (commented):
ok

PR #397, opened 2023-06-27
some little fixes, improvements and tests

noha:


MarcusDenker:
testAddToExistingEmptyList might be failing for BTree on the matrix test, we could remove the Btree for now from the setup, 

(SoilIndexDictionaryTest class>>#testParameters)

I will then have a look independedly (the Btree code is not yet working for real with the changes)

PR #416, opened 2023-07-04
added critical section for reading and writing database version

bocasti:
I wonder what would could be the result of concurrent reading and writing of the database version.
I don;t know if that's the cause, but it feels like there could be conflicts if two concurrent transaction in the same image try to read and write the setting file at the same time


noha:
@bocasti I've did #418 to solve this. Thanks but this is closed

PR #426, opened 2023-07-11
checkpoint and continue should both do nothing when recordsToCommit isEmpty

yannij:
Add guard clause to SoilTransaction>>#continue when recordsToCommit is empty, which parallels what happens in #checkpoint. Otherwise, the journal entry sometimes has a writeVersion of nil, leading to an error.

### DIFF
--- a/src/Soil-Core/SoilTransaction.class.st
+++ b/src/Soil-Core/SoilTransaction.class.st
@@ -234,6 +234,8 @@ SoilTransaction >> commit [
 
 { #category : #actions }
 SoilTransaction >> continue [
+	"recordsToCommit ifEmpty: [ self crTrace: 'EMPTY recordsToCommit' ]."
+	recordsToCommit ifEmpty: [ ^ self ].
 	recordsToCommit do: [ :record | | persistentRecord |
 		persistentRecord := record asPersistentClusterVersion.
 		objectMap at: record object put: persistentRecord.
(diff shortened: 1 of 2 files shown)


src/Soil-Core/SoilTransaction.class.st
```diff
@@ -234,6 +234,8 @@ SoilTransaction >> commit [
 
 { #category : #actions }
 SoilTransaction >> continue [
+	"recordsToCommit ifEmpty: [ self crTrace: 'EMPTY recordsToCommit' ]."
```
noha, line 237: If you do a PR you should remove commented out stuff before, please

yannij: Got it. Probably a bad habit. I should instead write more tests, and then trust the tests.

noha: It would still be good to remove it ;)

yannij: I removed the commented out code. I thought you meant for future pull requests. I thought pull requests could not be changed, after submission, but it makes sense that they're actually easy to update. (My usage of github has been limited)

noha: A pull request refers to a branch that you can change and github will redo the build. It would be hard otherwise what to do if the build fails ;)

noha (approved):

PR #432, opened 2023-07-20
#next: for index interator

### LINKED ISSUE #304: SoilIndexIterator: #next:
Inspectors on indexes and indexed collections need to only show parts of the data (indexes can get very large an disk and inspecting should not force load evertything)

For paged inspectors, we need a way to load the next n entries, the basis for that would be #next: on SoilIndexIterator.

MarcusDenker:
- add #SoilIndexIterator>>#next:
- add testIteratorFindAndNext2 testing next: for SkipList
- implement testIteratorFindAndNext and testIteratorFindAndNext2 for BTree

- improve testFirstWithTransaction to test first:

fixes #304

### DIFF
--- a/src/Soil-Core/SoilIndexIterator.class.st
+++ b/src/Soil-Core/SoilIndexIterator.class.st
@@ -106,6 +106,17 @@ SoilIndexIterator >> next [
 ]
 
 { #category : #accessing }
+SoilIndexIterator >> next: anInteger [
+	| result |
+	result := OrderedCollection new: anInteger.
+	anInteger timesRepeat: [
+		| next |
+		next := self next ifNil: [ ^ result].
+		result add: next ].
+	^ result
+]
+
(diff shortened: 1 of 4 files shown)


src/Soil-Core/SoilIndexIterator.class.st
```diff
@@ -105,6 +105,17 @@ SoilIndexIterator >> next [
 	^ self nextAssociation value
 ]
 
+{ #category : #accessing }
+SoilIndexIterator >> next: anInteger [
+	| result |
+	result := OrderedCollection new: anInteger.
+	anInteger timesRepeat: [
```
noha, line 112: If the remaining items are smaller then the requested anInteger is it intended that the resulting OrderedCollection has the size of anInteger with trailing nils instead of an OrderedCollection that has the size of the remaining items

noha (approved):

Issue #415, opened 2023-07-01
SoilSkipListDictionary #last and first: order newValues vs Order of the Index


MarcusDenker:
newValues is an OrderedDictionary, but this means that we add value in order of when we add.

But the Index datastructure is ordered according the key-order. e.g. if we use keys where we know the order and then add first the second, then the first, this test fails:

```
testFirst

	dict at: 2 put: #bar2.
	dict at: 1 put: #bar.

	self assert: dict first equals: #bar.
	self assert: (dict first: 1) first equals: #bar.
```

MarcusDenker:
The test with a transaction is ok:

```
testFirstWithTransaction
	| tx tx2 |
	tx := soil newTransaction.
	tx root: dict.
	dict at: 2 put: #two.
	dict at: 1 put: #one.
	
	self assert: dict first equals: #one.
	tx commit.
	"open a second transaction ..."
	tx2 := soil newTransaction.
	"and test first"
	self assert: tx2 root first equals: #one

```

MarcusDenker:
I think instead of 

```
newValue associations last
```

we need to create a new sorted collection that sorts in key-order. 

OrderedDictionary>>#associations now is:

```
associations
	| associations i |

	associations := Array new: self size.
	i := 1.
	self associationsDo: [:each |
		associations at: i put: each.
		i := i + 1].
	^ associations
```

that is, it returns the associations sorted by order of adding, we need an array sorted by key order

MarcusDenker:
instead of using "newValues associations", I think we could use a sorted version like this:

```
newValuesSortedByKeyOrder

	^ newValues associations sort: [ :a :b |
		(a key asSkipListKeyOfSize: self index keySize) asInteger 
		< (b key asSkipListKeyOfSize: self index keySize) asInteger ]
```

MarcusDenker:
PR is waiting for https://github.com/ApptiveGrid/Soil/pull/435 to be merged, as with git any change after a rename of a class will be a conflict

PR #438, opened 2023-07-24
415-SoilSkipListDictionary-last-and-first-order-newValues-vs-Order-of-the-Index 

MarcusDenker:
trying an idea for  #415 by using newValues sorted by key order (see  #testFirst for a test failing before but working with this change)

Note: in conflict with PR #435, so whichever is merged fist, the other has to be re-done (git does not understand class renames)


MarcusDenker:
will redo...

PR #457, opened 2023-08-12
Make serializer standalone

noha:


noha:
Pharo12 failing test is yet again UUID. Need to figure out the bytes for pharo12

PR #468, opened 2023-09-11
basic setup to flush cached object records and index pages

noha:


src/Soil-Core-Tests/SoilIndexTest.class.st
```diff
@@ -92,6 +92,35 @@ SoilIndexTest >> testConcurrentRemoveKey [
 
 ]
 
+{ #category : #tests }
+SoilIndexTest >> testFlushIndexPages [
+	| tx dict capacity txn2 txn1 root1 root2 |
+	tx := soil newTransaction.
+	dict := SoilSkipListDictionary new
+		keySize: 16;
```
MarcusDenker, line 100: I will (but after this is merged) check to make it to work for both BTree and SkipList

MarcusDenker (approved):
ok

PR #469, opened 2023-09-13
make transaction class extendable by method call or by subclassing

noha:


MarcusDenker (approved):

MarcusDenker:
test fail due to testRootObjectModifies uses #isChanged

PR #486, opened 2023-10-26
[WIP] Introduce journal fragments

noha:
This PR changes the way a journal is written. Right now for each transaction a file is created. This is not appropriate when having lots of transactions. Now the transactions are written to fragment files. Each can be 16MB of size which reduces the number of files significantly.


MarcusDenker (approved):

noha:
Merging into feature branch in order not too get a too big PR

Issue #459, opened 2023-08-14
Check if removed index items can be removed on read/write


noha:

MarcusDenker:
When removed index items are cleaned, we need to make sure:

- to update the index structures (index pages for Btree)
- merge pages if possible
- updating the index pages in case of BTree might lead to further merges (of index pages)

PR #501, opened 2023-11-13
Index: Cleanup low level unused removeKey:

### LINKED ISSUE #272: B+Tree: Removal needs to update index pages
Implement removal from a B+Tree

MarcusDenker:
The low level indexes (BTree and SkipList) can not implement a naive #removeKey:

#removeKey: is implemented on the level of IndexedDictionary and marks the entry as deleted. This can only be cleaned up later, one idea is to do it on save (see #459). At that point then we have to remove the deleted item, possibly merge pages and update the index.

This PR removes the #removeKey: implementation on Btree and SkipList, as it is very confusing
- it is not called outside fo the low level tests (not used by IndexDictionary and can not be used by it)
- does not update the indexes 

closes  #272

### DIFF
--- a/src/Soil-Core-Tests/SoilBTreeTest.class.st
+++ b/src/Soil-Core-Tests/SoilBTreeTest.class.st
@@ -312,21 +312,6 @@ SoilBTreeTest >> testPageAddFirstAndLoad [
 ]
 
 { #category : #tests }
-SoilBTreeTest >> testRemoveKey [
-
-	| capacity |
-	capacity := btree headerPage itemCapacity.
-	1 to: capacity do: [ :n | btree at: n put: #[ 1 2 3 4 5 6 7 8 ] ].
-	
-	btree removeKey: 20.
-	
-	self assert: (btree at: 1) equals: #[ 1 2 3 4 5 6 7 8 ].
-	self assert: (btree at: capacity) equals: #[ 1 2 3 4 5 6 7 8 ].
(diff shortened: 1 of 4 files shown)


noha:
But the indexes can actually do a removeKey: The key is just an integer that is managed by the SkipList/BTree. So I don't really understand the rationale here

MarcusDenker:
It is implemented right now just on SoilIndexedDictionary>>#removeKey:ifAbsent:, here it does not remove but instead it adds removed entry. 


```
removeKey: key ifAbsent: aBlock
	| iterator v |
	^ transaction 
		ifNotNil: [ 
			"remove from newValues as there could be a new at:put: on that
			key but removing the key will remove the value again"
			newValues removeKey: key ifAbsent: nil.
			transaction markDirty: self.
			iterator := self index newIterator.
			v := iterator at: key ifAbsent: [^ aBlock value ].
			v isRemoved ifTrue: [ ^ aBlock value ].
			removedValues 
				at: key 
				put: v asSoilObjectId.
			iterator at: key put: (SoilObjectId segment: 0 index: 0) ]
		ifNil: [ 
			removedValues 
				at: key
				put: (newValues removeKey: key ifAbsent: [ ^ aBlock value ]) ]
```

noha:
Yes I understand the use case in the IndexedDictionary case and also flagging removed values. But the Indexes are kind of dictionaries themselves where you should be able to remove an item, no?

MarcusDenker:
Ok, I close this

PR #503, opened 2023-11-13
BTree-Flush-Index 

MarcusDenker:
- testOverflowCopyOnWriteSplitting for BTree
- copy over testFlushIndexPages and fix Btree to work correctly


noha:
Hmmm, but the build was red

MarcusDenker:
As I merged this in a new image and then commited but the branch was wrong, it was merged somehow, Very strange

MarcusDenker:
And I can not revert

PR #518, opened 2023-11-23
Move #isRemoved for #isEmpty from the pages to the SoilIndexedDictionary

MarcusDenker:
The test of removed items should not happen at the level of the pages

- remove #hasItems
- use #isEmpty on the level of the SkipList and Btree, not #hasItems
- do the #isRemoved check on the level of the SoilIndexedDictionary

Clean up first/last item to not need a block creation.

This does not restore values yet, it is just a refactoring. But now we can use it in a next step

### DIFF
--- a/src/Soil-Core/SoilBTreePage.class.st
+++ b/src/Soil-Core/SoilBTreePage.class.st
@@ -49,7 +49,8 @@ SoilBTreePage >> find: aKey with: aBTree [
 
 { #category : #accessing }
 SoilBTreePage >> firstItem [
-	^ items first
+
+	^ items isNotEmpty ifTrue: [ items first ]
 ]
 
 { #category : #testing }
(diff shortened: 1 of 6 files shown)


src/Soil-Core/SoilBTreePage.class.st
```diff
@@ -49,7 +49,8 @@ SoilBTreePage >> find: aKey with: aBTree [
 
 { #category : #accessing }
 SoilBTreePage >> firstItem [
-	^ items first
+
+	^ items isNotEmpty ifTrue: [ items first ]
```
noha, line 53: what does it return when items is empty?

noha (approved):

src/Soil-Core/SoilBTreePage.class.st
MarcusDenker: nil,maybe it is better to write is explictly in an ifFalse:

PR #526, opened 2023-11-30
Check restored values for #isEmpty

### LINKED ISSUE #520: SoilIndexedDictionary>>#isEmpty and restored values
SoilIndexedDictionary>>#isEmpty does not restore, it just tests for isRemoved

MarcusDenker:
This PR retores the value for isEmpty

Fixes #520 

testConcurrentIsEmpty shows that #size does not restore values, I opened another issue for that


MarcusDenker:
I did another pass and implemented #isEmpty using SoilIndexedDictionary>>#restoreValue:forKey:iterator:

This is nice, as this way the only method on SoilIndexedDictionary that does an isRemoved check can be restoreValue:forKey:iterator: (together with the pr #530 for basicAt:ifAbsent:)

noha (approved):

PR #533, opened 2023-12-06
Fix indexed dictionary modificiations

noha:
current updates to an index could only be written when the indexed dictionary has modifications. In order to succeed the indexed dictionary was marked dirty when a key was added or removed to its index making it a central conflict point. 

Handling of modifications to the indexed dictionary itself and to its index are separated now. For this I introduced indexUpdates to be scanned on checkpoint and treated after the records to commit


noha:
The build still fails because there are tests that assume concurrent index modifications raise an error. But this was just a coincidence because the indexed dictionary was marked dirty failing for the same reason.

MarcusDenker (approved):

PR #544, opened 2023-12-08
SoilIndexedDictionary: implement size to restore values

### LINKED ISSUE #527: SoilIndexedDictionary>>#size and restored values
testConcurrentIsEmpty shows that #size takes into account that items where removed, but it does not restore values

MarcusDenker:
this uses the #do of the SoilIndexedDictionary in size, this way it is can restore values correctly on the level of the SoilIndexedDictionary.

fixes #527

### DIFF
--- a/src/Soil-Core/SoilIndexedDictionary.class.st
+++ b/src/Soil-Core/SoilIndexedDictionary.class.st
@@ -339,7 +339,12 @@ SoilIndexedDictionary >> second [
 { #category : #accessing }
 SoilIndexedDictionary >> size [ 
 	^ transaction 
-		ifNotNil: [ self index size ]
+		ifNotNil: [ 
+			"We should just call size on index, but the do: here can restore values"
+			| sum |
+			sum := 0.
+			self do: [ :each | sum := sum + 1 ].
+			sum ]
 		ifNil: [ newValues size ]
 ]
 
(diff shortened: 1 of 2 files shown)


src/Soil-Core/SoilIndexedDictionary.class.st
```diff
@@ -339,7 +339,12 @@ SoilIndexedDictionary >> second [
 { #category : #accessing }
 SoilIndexedDictionary >> size [ 
 	^ transaction 
-		ifNotNil: [ self index size ]
+		ifNotNil: [ 
+			"We should just call size on index, but the do: here can restore values"
+			| sum |
+			sum := 0.
+			self do: [ :each | sum := sum + 1 ].
```
noha, line 346: this does not scale at all. Not sure how we do it until now but doing a linear scan counting is not the way to go

noha (commented):
Let's discuss!

src/Soil-Core/SoilIndexedDictionary.class.st
MarcusDenker: But we do already do the same linear scam on the index...

MarcusDenker:
```
SoilBasicSkipList>>#size
	"We iterate over all elements to get the size. Slow!"
	^ self newIterator size 
```

PR #548, opened 2023-12-14
move value restore from the SoilIndexedDictionary to SoilIndexIterator

### LINKED ISSUE #545: Move handling of restored values to SoilIndexIterator
Right now, we deal with isRemoved in SoilIndexIterator, while restoring values from the transaction log is done only by SoilIndexedDictionary.

This is problematic, as *if* there is a transaction, just ignoring isRemoved is not enough. We always have to check the transaction log.

# How it works now

SoilIndexedDictionary>>#restoreValue:forKey:iterator:
	calling SoilIndexedDictionary>>#historicValueAt:iterator:ifAbsent:
		using SoilTransaction>>#journalEntriesFor:inIndex:startingAt:
		
But #journalEntriesFor:inIndex:startingAt: does not need the transaction, it need the readVersion *and* the soil instance to read the transaction log.

# Can we move this to SoilIndexIterator?

If we move this to SoilIndexIterator, this means we need to have readVersion and soil.

But the iterator is created and used by the index (SoilBasicBTree or SoilBasicSkipList) via #newIterator

```
SoilBasicSkipList>>#newIterator
```

This means, we would need to have these classes know readVersion and soil, too, when created for the SoilIndexedDictionary

```
SoilSkipListDictionary>>#createIndex
	^ SoilSkipList new
		initializeHeaderPage;
		valueSize: 8;
		yourself
```

# Open Question: Standalone usage and isRemoved without restore

There are usages of the indexes without SoilIndexedDictionary, See SoilBehaviorRegistry>>#initializeFilesystem 

- Do they have to deal with restoring values? No, as this usage never adds a Removed OID to the index.
- This means without soil / readVersion, the iterator does not need to check for isRemoved for this use
- What about Visitors without transaction (see https://github.com/ApptiveGrid/Soil/issues/532)

MarcusDenker:
fixes #545

### DIFF
--- a/src/Soil-Core/SoilBTreeDictionary.class.st
+++ b/src/Soil-Core/SoilBTreeDictionary.class.st
@@ -9,6 +9,8 @@ SoilBTreeDictionary >> createIndex [
 	^ SoilBTree new
 		initializeHeaderPage;
 		valueSize: 8;
+		soil: (transaction ifNotNil: [:tr | tr soil]);
+		readVersion: (transaction ifNotNil: [:tr | tr readVersion]);
 		yourself
 ]
 
--- a/src/Soil-Core/SoilBasicBTree.class.st
+++ b/src/Soil-Core/SoilBasicBTree.class.st
@@ -2,7 +2,9 @@ Class {
 	#name : #SoilBasicBTree,
 	#superclass : #Object,
 	#instVars : [
-		'store'
+		'store',
+		'soil',
+		'readVersion'
 	],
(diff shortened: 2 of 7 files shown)


MarcusDenker:
This moves the restore value code to SoilIndexIterator as #restoreValue:forKey.

As before, it is called in do:, at:ifAbsent: and isEmpty, but not yet in #nextAssociation

TODO
- instead of calling in in do, we need to do call it in nextAssociation instead. 
- check user of nextPresent and nextPresentAssociation

PR #569, opened 2024-01-15
523-Index-Iterator-firstAssociation--need-to-restore-values 

### LINKED ISSUE #523: Index Iterator: #firstAssociation need to restore values
After merging #522, do: is not using #isRemoved. But the method that is uses (nextPresent) does not restore values, either

MarcusDenker:
This PR
- adds testFirstAssociationWithTransactionRemoved which tests restoring if removed values for #firstAssociation
- implement #firstAssociation in terms of #nextAssociation without the need to check #isRemoved here

fixes #523


MarcusDenker:
Pharo 12 fails due to https://github.com/pharo-project/pharo/issues/15953

noha (approved):

PR #575, opened 2024-01-17
  571-SoilIndexedDictionarynextAfter-should-forward-to-iterator 

### LINKED ISSUE #571: SoilIndexedDictionary>>#nextAfter: should forward to iterator
To follow the pattern of all the other methods, we should implement

SoilIndexedDictionary>>#nextAfter:

to foward to the iterator and have the implementation there. This then allows us to do the same from the low level index, too and have a unified API

MarcusDenker:
- fix testNextAfter, we should return the association for both the case with and without a transaction

- add #nextAssociationAfter:  to SoilIndexIterator

I added a comment to #nextAfter: that it returns the association. Maybe it should be renamed to be nextAssociationAfter:

fixes #571


noha:
I'm for the rename

MarcusDenker:
rename is done

noha (approved):

PR #600, opened 2024-02-09
#asIndexKeyOfSize: cleanup

MarcusDenker:
Instead of having

```
 (aString asIndexKeyOfSize: self keySize) asInteger
```

everywhere, this PR moves #binaryKey: to the index. We can use that to calculate the key with the right length


noha:
Wouldn't it be better to call #binaryKey rather #indexKey because it is the key in the index? 
Furthermore #asIndexKeyOfSize should probably return the integer directly.

MarcusDenker:
I made an improved version

PR #601, opened 2024-02-09
remove nextKeyCloseTo: on the level of  SoilIndexedDictionary

### LINKED ISSUE #596: revisit SoilIndexedDictionary>>#nextKeyCloseTo:
#nextKeyCloseTo: allows us to lookup in a dictionary for the next entry even if they key itself is not there (#nextAfter: would raise an error).

#nextKeyCloseTo: operates for that on the the indexKey, (the #asIndexKeyOfSize:) and returns the next one. We could use this to do another lookup to get the value.

The question is: does it make sense to have this method on the level of SoilIndexedDictionary? The users there have no use for the indexKey, the only thing they could do is to look up a values and get confused between the key they uses as an argument and the indexKey.

-> remove #nextKeyCloseTo: from SoilIndexedDictionary
-> add #nextCloseTo: which would return the value (like #nextAfter: )

MarcusDenker:
this removes nextKeyCloseTo: from the SoilIndexedDictionary now that we have #nextCloseTo:

This is was last API on SoilIndexedDictionary that exposed indexKeys.

fixes #596

### DIFF
--- a/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
+++ b/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
@@ -533,21 +533,6 @@ SoilIndexedDictionaryTest >> testNextAssociationAfterWithTransaction [
 ]
 
 { #category : #tests }
-SoilIndexedDictionaryTest >> testNextKeyCloseToWithTransaction [
-	| tx tx2 |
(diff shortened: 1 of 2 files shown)


MarcusDenker:
argh.. wrong merge. I will redo

PR #606, opened 2024-02-09
this PR makes sure to always use #proxyForObjectId:

### LINKED ISSUE #603: SoilIndexedDictionary: #proxyFromByteArray: vs  #proxyForObjectId:
In SoilIndexedDictionary, some users use   #proxyForObjectId:, others : #proxyFromByteArray: 

The difference is that : #proxyFromByteArray:  sends an additional #asSoilObjectId, thus it works for both cases.

MarcusDenker:
in the SoilIndexedDictionary (which sends #asSoilObjectId)

fixes #603


MarcusDenker:
proxyForObjectId: does send  #asSoilObjectId, too

So we can just use proxyForObjectId:, I will update the PR

noha (approved):

PR #608, opened 2024-02-13
Change stream of fragment files to be a lockable stream. When writing…

noha:
… to the fragment file lock it before


MarcusDenker (commented):
ok

PR #616, opened 2024-02-15
Small cleanup: fix #halt SoilIndexConsistencyVisitor 

MarcusDenker:
- SoilIndexConsistencyVisitor better uses Error instead of halt (like SoilJournalConsistencyVisitor)
- make sure to have the error be descriptive

This can be improved furhter later as needed (e.g. at some point we could introduce special errors)


MarcusDenker:
Smalltalk CI is broken for Pharo12

MarcusDenker:
(PR was done while reading the code, I will in another step update SoilIndexConsistencyVisitor to support both SkipList and Btree and use it to check BTree consistency)

noha (approved):

PR #621, opened 2024-02-16
ignore removed items when reading an items page

noha:


MarcusDenker (approved):

MarcusDenker:
For the SkipList this works, for B+Tree I need to check how to update the index pages, as they are now structured as if the removed value would still be there.

noha:
we can postpone this a bit

PR #623, opened 2024-02-16
added pointer for last free page index

noha:


noha:
@MarcusDenker I implemented lastFreePageIndex only in the skip list header page. No idea how that applies to BTree

MarcusDenker (approved):
Ok, I will check how to apply that to the BTree

PR #627, opened 2024-02-17
626-SoilBTreeHeaderPagereadFrom--should-call-super

### LINKED ISSUE #626: SoilBTreeHeaderPage>>#readFrom:  should call super
SoilBTreeHeaderPage>>#readFrom:  copies the code instead of calling super

the class structure should be changed to have Header and Data page be subclasses of a common superclass

MarcusDenker:
add SoilAbstractBTreeDataPage, this way the two subclasses can implemnent reading/writing in a simpler way (and re-use the code of SoilIndexPage)

fixes #626


MarcusDenker:
I added the needed changes for the lastFreePageIndex to this PR, too (for the Data pages, for the index page I will do another PR later)

PR #628, opened 2024-02-17
Optimized  #isRemoved

MarcusDenker:
avoid #asInteger for #isRemoved when the second byte is not 0 (#at: is very fast, while #asInteger is slow)

### DIFF
--- a/src/Soil-Core/ByteArray.extension.st
+++ b/src/Soil-Core/ByteArray.extension.st
@@ -17,7 +17,7 @@ ByteArray >> asSoilObjectProxy [
 
 { #category : #'*Soil-Core' }
 ByteArray >> isRemoved [
-	"First two bits are the segment ID, for normal objects it is 1, 
+	"First two bytes are the segment ID, for normal objects it is 1, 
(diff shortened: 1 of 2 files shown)


noha (commented):
The segment byte size should configurable later so we should not forget. Maybe a flag might do

noha:
The question is why is asInteger slow and if so shouldn't we store integers instead?

MarcusDenker:
I added a comment to #newFrom:

Using integers is an interesting idea... but these will be large integers  (as we use 64bit). Which are just wrapped ByteArrays, but I wonder if it is then really that much faster

MarcusDenker:
Ah, dadurch das die segment size immer sehr klein ist, passen alle IDs die wir zur zeit verwenden in einen SmallInteger.

D.h. der fall das es ein LargeInteger wird waere nur für andere verwendungen der indexes relevant (die tests z.b. verwenden auch andere grössen fuer die values).

Ist also etwas wo es mal interesstant wäere, benchmarks zu machen.

PR #644, opened 2024-02-22
Update documention: Pharo12 class deprecatedAlias support

MarcusDenker:
.. it just works.

I will add a test, in the long term we could remove the soil specific class rename feature


noha (approved):

noha:
But is the deprecated stuff stored along the code?

MarcusDenker:
yes, it is normally in a class side initialize method now. Later I want to have this as meta-data in the class definition

PR #645, opened 2024-02-22
testing class rename support using aliases on Pharo12

MarcusDenker:
.. this is nice that it just works


MarcusDenker:
AARGH but now there are so many changes again due to commiting from pharo12

I close this and check that, I think we can now fix the version of tonnel used or something?

PR #647, opened 2024-02-24
Replay journal on open

noha:
Adds a journal recovery operation which is executed every time a soil instance is opened. This way we should be sure that the heap and/or the journal are in consistent state


noha:
I need to fix the backup still. The problem is that an initial journal fragment file is created to early and prevents overwriting that

noha:
Fixed it. But we need to separate the ways of how we initialize a database. For database creation we need to create initial files like the first journal fragment file. For backup mode we should not create those as they are copied over from another instance. Maybe we need to add something like initializeEmpty for not creating any additional things

MarcusDenker (approved):
ok

PR #648, opened 2024-02-26
separated iterator between a raw iterator that acts raw on the index …

noha:
…structure and a index iterator that has a transaction converting values


src/Soil-Core/SoilIndexIterator.class.st
```diff
@@ -279,9 +231,9 @@ SoilIndexIterator >> nextAssociation [
 	| nextAssociation nextValue key |
 	nextAssociation := self basicNextAssociation ifNil: [ ^nil ].
 	key := nextAssociation key.
-	nextValue := self restoreItem: nextAssociation.
+	nextValue := self convertItem: nextAssociation.
```
MarcusDenker, line 234: here "convertedItem" would be better then "nextValue" as a name

MarcusDenker (commented):
ok

PR #655, opened 2024-03-02
[WIP] Proposal for storing database wide metadata

noha:
This is a proposal for adding a extensible but static format for storing data outside of a transaction.

The rationale is have a file that can store a dictionary of values. The values of the dictionary can be all types that have a type code, meaning all that are covered by the basic serializer/materializer. These don't need a behavior registry and do not interfer with the inner workings of the database.

The format is extensible as it is a method that defines it like: 

```
format 
	^ { 
		{ #lastOpened . [ DateAndTime now ] . 33  } .
		{ #lastModified . [ DateAndTime now ] . 33  } .
		{ #foo . [ 0 ] . 8 } }


```

The definition is:

- the name of the property to store (here lastOpened, etc.)
- a block that returns the initial value for the property and therefor the type of the value
- the fixed size needed for the storage of the value. In the example above this is 33 for DateAndTime (32 contents + 1 size byte), 8 for Integer (this is written via length encoded integer so no overhead). The resulting storage sizes computed are 34 for DateAndTime (33 + 1 type code) and 9 for Integer (8 for contents and 1 type code)

As long as we make the class registerable a user of the database can created a subclass and register that which allows to have custom formats per use case

### DIFF
--- a/src/Soil-Core/Soil.class.st
+++ b/src/Soil-Core/Soil.class.st
@@ -38,7 +38,8 @@ Class {
 		'serializerClass',
 		'materializerClass',
 		'control',
-		'setup'
+		'setup',
+		'metadata'
 	],
 	#category : #'Soil-Core-Model'
 }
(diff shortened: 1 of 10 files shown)


noha:
The serialization of DateAndTime to string is just a short cut for now. This is likely to be changed

MarcusDenker (approved):

PR #656, opened 2024-03-02
Changed handling of bytes to convert before and use nextPutAll: as ne…

noha:
…xtPutBytesFrom: is a fuel method and should not be used.

Moved nextSoilObject to basic materializer as it makes use of it

### DIFF
--- a/src/Soil-File/SoilLockableStream.class.st
+++ b/src/Soil-File/SoilLockableStream.class.st
@@ -146,6 +146,11 @@ SoilLockableStream >> position: anInteger [
 	fileStream position: anInteger 
 ]
 
+{ #category : #accessing }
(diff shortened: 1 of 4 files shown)


noha:
@MarcusDenker There is now a test failing and I'm not sure why. Can you have a look?

MarcusDenker:
The problem is that we need. way to store *all* ByteLayout objects. They might not implement #asByteArray.
#nextPutAll: can not use non-bytearrays

This is of course another case of "I just add my own method instead of fixing the system"... Fuel should have fixed #nextPutAll: or add #nextPutBytesAll: as as sytem method. But everyone just "builds on top" instead of improving the base for everyone.

I suggest to add for now a copy of nextPutBytesFrom: to Soil.

noha:
I reverted the change and make a ticket

PR #658, opened 2024-03-03
Add testIndexAfterReopen which fails in the scenario because header page has pageSize nil

yannij:
When a previously committed BTree or SkipList is accessed in the test scenario, the header page object has a pageSize of nil. The scenario uses a freshly opened Soil, then requests all the dictionary values, before using #at:put:. The commit then triggers a DNU where the pageSize is nil in the #hasRoom method.

==== stack trace ====
UndefinedObject(Object)>>doesNotUnderstand: #-
UndefinedObject>>doesNotUnderstand: #-
SoilSkipListHeaderPage(SoilIndexItemsPage)>>hasRoom
SoilSkipListIterator>>basicAt:put:
SoilSkipListIterator(SoilIndexIterator)>>at:put:
SoilSkipListDictionary(SoilIndexedDictionary)>>at:put:
SoilIndexedDictionaryTest>>testAfterReopen

### DIFF
--- a/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
+++ b/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
@@ -34,6 +34,13 @@ SoilIndexedDictionaryTest >> path [
 	^ 'soil-tests'
 ]
 
+{ #category : #tests }
+SoilIndexedDictionaryTest >> reopen [ 
+	soil ifNotNil: [ 
+		soil close ].
+	soil open
+]
+
 { #category : #running }
 SoilIndexedDictionaryTest >> setUp [ 
 	super setUp.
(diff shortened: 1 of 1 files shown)


MarcusDenker:
I think we need to set the page size: in SoilPagedFileIndexStore>>#pageFaultAt:

It is not stored as we need the size to read, but we never set it.

```
pageFaultAt: anInteger
	| page |
	self ensureStreamIsOpen.
	streamSemaphore critical: [  
		stream position: (self positionOfPageIndex: anInteger).
		page := (index readPageFrom: stream) 
			index: anInteger;
			pageSize:  self filePageSize
			yourself ].
	pagesMutex critical: [  
		pages at: anInteger put: page ].
	^ page
```

MarcusDenker:
other PR is merged. Thanks!

noha:
@MarcusDenker btw. with PRs from collaborators. You can make a new integration branch, change the PR of the collaborator to merge in that integration branch, add your changes and then make a PR with the combined change. This way the history is easier to follow

PR #663, opened 2024-03-04
Added complete sync of database files

noha:
This PR improves the usage of terms commit and checkpoint in Soil. 

- A commit is the generation of a transaction journal and committing this to the heap. In this stage the journal is written to disk using fdatasync(), the data to the heap is just written and flushed which hands out the data to the OS but does are not on disk. 
- A checkpoint makes sure that all heap files are written to disk and it generates a checkpoint entry in the journal to mark up to which point the data is persisted in a guaranteed way.

The changes to improve this:

- the selector checkpoint in the control file has been renamed to lastCheckpoint which makes reading about the checkpoint easier
- the checkpoint method in the transaction was renamed to basicCommit. Before the checkpoint method was doing a commit plus checkpoint. Now it only does the commit phase. In order to be backward compatible the #commit selector does basicCommit and checkpoint which is a preparation to separate both completely
- there is now a checkpoint method in Soil which moved up from the persistent journal. It creates an checkpoint entry, the entry being committed will write all heap files to disk and after the checkpoint entry is written to the journal

NOTE: Soil will be a lot slower now. Before there was no assurance that files are written properly. Now with the recovery we close it in a way it should be safe. But for the same reason it will be slow. Especially as we still do a checkpoint per transaction commit it will always force writing and that makes it slow. Later we move the checkpoint to a later time which accelerates the committing

### DIFF
--- a/src/Soil-Core-Tests/SoilDatabaseJournalTest.class.st
+++ b/src/Soil-Core-Tests/SoilDatabaseJournalTest.class.st
@@ -93,7 +93,7 @@ SoilDatabaseJournalTest >> tearDown [
 SoilDatabaseJournalTest >> testRecoverWithAtEnd [
 	soil close.
 	soil open.
-	self assert: soil control checkpoint equals: (SoilLogSequenceNumber fileNumber: 0 offset: 28) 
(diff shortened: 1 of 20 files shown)


MarcusDenker (approved):
ok!

PR #684, opened 2024-04-22
Integrate fix cyclic references

noha:


noha:
I close this for now. There is an issue for the problem

PR #693, opened 2024-05-16
Update build.yml: add Pharo13

MarcusDenker:
this adds a ci run for Pharo13


MarcusDenker:
I added that change to https://github.com/ApptiveGrid/Soil/pull/694

Issue #690, opened 2024-05-13
For Pharo11/Pharo12 parallel use: fix format to tonel 1.0 for now


MarcusDenker:
We should I think for now fix the format used to Tonel 1.0, this way we can use both Pharo11 and Pharo12 in parallel to contribute.

Form the mailinglist:

If you want to convert all the files of a repository at once to avoid to have multiple PR with format changes you can use this script and commit the resulting files:

```
| projectName |
projectName := 'Spec2'.
repository := IceRepository repositories detect: [ :repo | repo name = projectName ].
repository workingCopy packages do: [ :pkg |
IceLibgitTonelWriter forInternalStoreFileOut: pkg latestVersion mcVersion on: repository ]
```

Also, if you work on a project both in p12 and p11, you can avoid the ping pong by fixing a version of tonel in the properties file. The file to update is the .properties that is in the source folder and it should look like this:

```
{
	#format : #tonel,
	#version: #'1.0'
}
```

P12 will take into account the tonel version asked and use it. In previous version of Pharo, this additional line will be ignored and Tonel v1 will be used.

PR #694, opened 2024-05-16
Pharo13: fix failing tests

MarcusDenker:
- the SystemVersion test to distingish 11 from 12 and later was wrong
- in Pharo12, we have some deprecation rewrites in the same code

This PR adds dedicated versions for Pharo12 (and later) and 11, this avoids the rewrite and makes it work in Pharo13


MarcusDenker:
Note: the PR now does the Tonnel conversion for all the classes it touches 

Maybe we should better do https://github.com/ApptiveGrid/Soil/issues/690 before

noha (approved):

PR #705, opened 2024-06-03
Enable restartig the journal on missing journal files instead of bloc…

noha:
…king the database from opening


MarcusDenker (commented):
ok

PR #718, opened 2024-06-11
Fixed basicNextAssociation:

noha:
- if the currentPage was empty it returned nil ending the search to early
- there can be an empty page at the end so it is possible to reach outside the loop. In that case returning nil is obvious


noha:
@MarcusDenker This PR fixes a central piece of the iterator which I want to carefully review. On the other hand I don't understand how that could have slipped through

MarcusDenker (commented):
I think we never saw that case as we never really removed items, but only set the value to isRemoved. 

During construction, empty pages never happen. Only when we remove items for real it can happen that we have an empty page that has a next page with values.

noha:
Same conclusion here. And I was right being cautious with removing all removed values on page load

PR #720, opened 2024-06-11
Added support to natively remove items from an index

noha:


MarcusDenker (commented):
For removing, we need to see how to update the index. Just using #removeItemsSuchThat: I think would lead to a bad state. 

Else good, I will write tests / see how to use SoilPluggableIndexRewriter for the BTree case

noha:
It is supposed to be used when you know what your doing. Apart from that I fail to see the bad state that is left behind.

PR #723, opened 2024-06-14
test and fix for #priorAssociation and #lastAssociation

MarcusDenker:
This PR adds a test for #priorAssociation where we call it exactly when we have to go back one page, thus needed to move to the prior page.

-> fixes a bug where we where checking for last page instead of first page


MarcusDenker:
Pharo 13 error is not due to this PR

noha (approved):

PR #730, opened 2024-06-19
We have #first: and #next:. but no #last:

MarcusDenker:
This PR add #last: on the iterator and the indexed dictionary and adds tests


MarcusDenker:
as I merged with main, github now shows all the changes already merged, too

noha (approved):

PR #758, opened 2024-06-26
unify #removeKey:, add tests, do not search index twice fro remove

MarcusDenker:
- add more tests, especially regarding return value
- test what happend if we remove already removed keys

This fixes SoilIndexIterator to return the value, the lookup fixes the behavior, too, for removing keys that are already removed (raises now an exception)

In addition, this fixes the double lookup that we had for removeKey: (was a implemented as #at: plus #at:put:), now we search only once and update the existing item if it was there.

### DIFF
--- a/src/Soil-Core/SoilIndexIterator.class.st
+++ b/src/Soil-Core/SoilIndexIterator.class.st
@@ -400,8 +400,15 @@ SoilIndexIterator >> removeItemsSuchThat: aBlock [
 
 { #category : #removing }
 SoilIndexIterator >> removeKey: key [
+	^ self 
+		removeKey: key 
+		ifAbsent: [ KeyNotFound signalFor: key in: self ]
+]
+
+{ #category : #removing }
+SoilIndexIterator >> removeKey: key ifAbsent: aBlock [
 	| returnValue |
-	returnValue := self at: key.
+	returnValue := self at: key ifAbsent: aBlock.
 	self 
 		at: key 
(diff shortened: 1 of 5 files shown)


MarcusDenker:
Need to check: we should not iterate twice for removeKey:

src/Soil-Core/SoilIndexIterator.class.st
```diff
@@ -400,8 +400,15 @@ SoilIndexIterator >> removeItemsSuchThat: aBlock [
...
+	^ self 
+		removeKey: key 
+		ifAbsent: [ KeyNotFound signalFor: key in: self ]
+]
+
+{ #category : #removing }
+SoilIndexIterator >> removeKey: key ifAbsent: aBlock [
 	| returnValue |
-	returnValue := self at: key.
+	returnValue := self at: key ifAbsent: aBlock.
```
noha, line 411: Doing two times #at: is expensive.  We should have that picking up at the current position

MarcusDenker: yes, we did this already, this pr moves the check now to the iterator (this at least avoids instantiating it twice)

But we have to do this in one step for sure, I will check how to do this

MarcusDenker: I have implemented SoilIndexIterator>>#removeKey:ifAbsent: to search only once

noha (approved):

PR #784, opened 2024-07-18
fixed time serialization to include nano seconds

noha:


MarcusDenker (approved):

MarcusDenker:
How do we deal with reading old data from disk ?

noha:
I just thought it might not be likely someone did it already so we get away with it 😉

PR #805, opened 2024-08-19
799-Iterator-basicAssociationsDo-is-not-basic

### LINKED ISSUE #799: Iterator: #basicAssociationsDo: is not basic ?
#basicAssociationsDo: seems to be not correctly named.

- #basicNextAssociation and #basicPreviousAssociation do not restore and convert
- #basicAssociationsDo: does restore and convert just as #do: does
- but the key is not converted

MarcusDenker:
- rename #basicAssociationsDo: to #associationsDo:
- implement #associationsDo: in term of #nextAssociation, as we want to restore the value *and* need associations anyway, no need to use #basicAssocations here

fixes #799

### DIFF
--- a/src/Soil-Core-Tests/SoilIndexIteratorTest.class.st
+++ b/src/Soil-Core-Tests/SoilIndexIteratorTest.class.st
@@ -78,6 +78,24 @@ SoilIndexIteratorTest >> testAddRandom [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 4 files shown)


MarcusDenker:
SmalltalkCI is broken in Pharo13, there is an issue already:https://github.com/pharo-project/pharo/issues/17012

noha (approved):

PR #818, opened 2024-09-19
add a hook to allow users to transform the materialized object

estebanlm:
useful, e.g. when restoring singletons.

this cannot be less overengineered, I think ;)


estebanlm:
I am green now :P 
now, if that's not the correct place, it certainly looks like (from my small understanding). At least my usecase works like a charm now :P

MarcusDenker (approved):

noha:
Ok, I need to do one more cycle on it because that should be possible for serialization and for materialization

noha:
I add it to an integration branch so can add missing stuff

PR #830, opened 2024-10-08
Create CONTRIBUTORS

MarcusDenker:
add a CONTRIBUTORS file, this lists all contributors

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,0 +1,6 @@
+Norbert Hartl <norbert@hartl.name>
+Marcus Denker <marcus.denker@inria.fr>
+Yanni Chiu https://github.com/yannij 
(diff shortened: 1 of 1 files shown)


MarcusDenker:
I was thinking of destinguishing main authors from those who just did some PRs.
(everyone listed in AUTHORS will be listed as AUTHOR on hal.sciences).

We can alternatively add everyone to AUTHORS, then have them as Authors in the HAL entry, too.

(For Pharo, I will suggest that we add everyone to AUTHORS who did a PR for that version, while CONTRIBUTORS will be the place where those stay even though the did not contribute to the current version)

noha:
Keep it this way. I just don't like this static information which will be mostly wrong because it needs manually work

PR #839, opened 2024-10-19
Recycle pages into free page list 

noha:


noha:
This PR goes on top of maintain-page-removal in order not to have a huge PR. This one needs cleanup and correct placement for btree to implement. Next is not only to manage free pages but to reuse

MarcusDenker (approved):

PR #856, opened 2024-10-29
Cleanup: move methods up in hierachy of Index classes

MarcusDenker:
Cleanup after page recycling was added to BTree (move methods / ivars up in hierachy, categorize...)


noha (approved):
Hooray! Great cleanup! Thanks!!

PR #869, opened 2024-11-28
Guard SoilPagedIndexStore>>#pageAt:put:  against overrides

MarcusDenker:
Make sure to never just override an exisiting page in the store.

We change SoilPagedIndexStore>>#pageAt:put: to use #at:ifPresent:ifAbsentPut: to raise an error in this case


MarcusDenker:
(I close the Pr)

PR #877, opened 2024-12-07
when removing a page add the previous page of the chain to dirty page…

noha:
…s so that its state gets written to disk


MarcusDenker (commented):
ok

PR #903, opened 2025-06-20
[Trivial] More tests, code coverage and duplicate code cleanup

MarcusDenker:
- add some more tests to improve code coverage
- move #allocatePage up in the SoilIndex hierarchy 
- tag some methods <ignoreForCoverage>


src/Soil-Serializer-Tests/SoilPrimitiveSerializationTest.class.st
```diff
@@ -225,7 +225,7 @@ SoilPrimitiveSerializationTest >> testSerializationDateAndTime [
 	self assert: serialized equals: #[29 5 10 0 0 5 0].
 	self assert: (serialized at: 1) equals: TypeCodeDateAndTime.
 
-	materialized := SoilBasicMaterializer materializeFromBytes: serialized.
+	materialized := SoilMaterializer materializeFromBytes: serialized.
```
noha, line 228: The tests in this class are about the basic materializer. We have the basic materializer that can handle primitives while the full materializer can handle graphs. The basic materializer is used e.g. to read the database configuration which is plain

PR #914, opened 2025-07-31
use slots to calculate size in memory

noha:


noha:
Not sure we need a PhD about it 😉 This is only used by persitent cluster to measure how much is cached in memory. So only usable if it is certain there are no loops etc.

PR #960, opened 2025-10-15
Add exporter

noha:


noha:
@MarcusDenker This the already reviewed PR that got lost due to rollback yesterday

PR #964, opened 2025-10-28
BTree backlinks

MarcusDenker:
add backlink ("previous") to BTree data pages and use it in #findPreviousPageOf:


MarcusDenker:
later I think we can refactor to have #findPreviousPageOf: to be the code that searches without backlinks (with a coment sayting that it will be removed eventually). But for now this is the minimal change to make it work

PR #976, opened 2025-11-14
Index: Document #removeKey: and empty pages

### LINKED ISSUE #850: Check SoilSkipListTest>>#testRemoveAllFromPage
SoilSkipListTest>>#testRemoveAllFromPage removes all items from a page, which we do not support (as we need a lowest key to search).

The test is green as SoilIndexIterator>>#basicNextAssociation has code to deal with empty pages

MarcusDenker:
This PR adds documentation to the index #removeKey: methods to explain that if the page gets empty, the index will not work anymore.

Maybe there is a way to have a basic remove and a remove: that would clean the page, but as Soil does not need that (but it needs instead control over page deletion for page reuse), I do not want to add that if we do not need it. 

fixes #850

### DIFF
--- a/src/Soil-Core/SoilBasicBTree.class.st
+++ b/src/Soil-Core/SoilBasicBTree.class.st
@@ -135,6 +135,8 @@ SoilBasicBTree >> pageClass [
 
 { #category : #removing }
 SoilBasicBTree >> removeKey: key ifAbsent: aBlock [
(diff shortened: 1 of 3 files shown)


noha:
Ok, but I think we should review that. I do not recall why the check of this is not in the index itself. The index has knowledge that there are pages and it holds the dirty pages anyway. Seems like an old workaround that could be improved

PR #982, opened 2025-11-17
Refactor platform sync

noha:
This PR moves the platform specific code to OSPlatform subclasses and uses them with a single dispatch


noha:
@mumez Can you review that as well? I think this way it is a bit cleaner

MarcusDenker (approved):

mumez (approved):

PR #1017, opened 2025-12-12
Iterator with duplicate keys: #removeKey: should remove all

### LINKED ISSUE #1014: Index with duplicate keys: #removeKey: should remove all
Situation right now for #removeKey: with duplicated keys:
- Iterator raises error
- index removes randomly the first it finds (and and this might actually break getting all the rest after) (see #988)

Instead of raising an error on the level of the index, we should instead on both the iterator and index implement #removeKey: to remove all the entries with that key.

MarcusDenker:
The PR implements on the level of the iterator to remove all key-value pairs on removeKey: when in duplicate key mode.

(after this is merged, we can see if we can turn the SoillInteratorTest into a 4-way Matrix test uniqkeys yes/no and BTree/SkipList

closes  #1014 (as there is another issue for removeKey: on the index level)

### DIFF
--- a/src/Soil-Core/SoilIndexIterator.class.st
+++ b/src/Soil-Core/SoilIndexIterator.class.st
@@ -583,7 +583,7 @@ SoilIndexIterator >> removeKey: key value: value ifAbsent: aBlock [
 SoilIndexIterator >> removeKeyAll: key ifAbsent: aBlock [
 	"we remove all the entries for key"
 	| all |
-	all := self at: key.
+	all := index at: key ifAbsent: aBlock.
 	all do: [ :each | self removeKey: key value: each ].
 	^all
 ]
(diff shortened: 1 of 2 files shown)


src/Soil-Core/SoilIndexIterator.class.st
```diff
@@ -579,6 +579,15 @@ SoilIndexIterator >> removeKey: key value: value ifAbsent: aBlock [
 	^ oldValue
 ]
 
+{ #category : #removing }
+SoilIndexIterator >> removeKeyAll: key ifAbsent: aBlock [
+	"we remove all the entries for key"
```
noha: as aBlock is not used it pretty much shows it is not needed ;) If we remove all keys we can just ignore if none is there. The declaration of having no key after is met

PR #1027, opened 2025-12-18
Update README.md: remove main warning

MarcusDenker:
Small readme update:

In the end, we kept main compatible to v3.

- remove warning
- add Pharo14


noha:
I would change the readme when we release

Issue #1040, opened 2026-01-08
Speedup Serializer for classes with no soilTransientInstVars


MarcusDenker:
We can use the same idea as  #1039 to speedup serialization for those classes that have no soilTransientInstVars.

Experiment: Change PointerLayout>>#soilBasicSerialize:with:

```
soilBasicSerialize: anObject with: serializer
	| description class |
	description := serializer serializeBehaviorDescriptionFor: anObject.
	class := anObject class.

	(class soilTransientInstVars size == 0) 
		ifTrue: [1 to: class instSize do: [:index | (anObject instVarAt: index)  theNonSoilProxy soilSerialize: serializer] ]
		ifFalse: [description instVarNames do: [:ivarName | ((anObject class slotNamed: ivarName) read: anObject) theNonSoilProxy soilSerialize: serializer ] ]
```

Benchmark Pharo12 without:

```
"#(#('ston' 602 1250 '9.45 MB') #('soil' 225 180 '1.60 MB') #('fuel' 1034 53"
```

with optimization:

```
"#(#('ston' 655 1299 '9.45 MB') #('soil' 162 185 '1.60 MB') #('fuel' 1035 53"

```

~40 % speedup.

PR #1041, opened 2026-01-08
Speedup serializer for classes with no soilTransientInstVars

MarcusDenker:
When there are no soilTransientInstVars, read the ivars per offset. fxes #1040


noha:
Never mind. I checked again and the check is just a less invasive way figuring out

PR #1054, opened 2026-01-19
Duplicate key restore: restore keys added and removed

MarcusDenker:
This PR changes restoreItem: to take the value into account for multiple keys.

The remove case looks incomplete, we definitly need more tests (e.g removing just some duplicate keys)


MarcusDenker:
Now I added correct restore for removed entries

- SoilRemoveKeyEntry now stores the removed entry
- This way do not need to treat remove specially im the restore logic
- the adding-after-remove case from above now works!

PR #1065, opened 2026-01-28
1061-Cleanup-needed-addItem-addItemDuplicate 

### LINKED ISSUE #1061: Cleanup needed: addItem: addItemDuplicate:
SoilSkipListPage>>#addItem: overrides SoilIndexItemsPage>>#addItem: with a change only needed for duplicate keys

We should check to unify this with SoilAbstractBTreeDataPage>>#addItemDuplicate:

- have both  SoilIndexItemsPage>>#addItem:  and  SoilIndexItemsPage>>#addItemDuplicate:,
-  change SkipList to use #addItemDuplicate:

MarcusDenker:
Simplify the addItem: code. Following the SkipList, we

- add a addItem: that is *only* for adding, supporting multiple keys
- change BTree to use this instead of having both addItemDuplicate: and addItem:

fixes #1061

### DIFF
--- a/src/Soil-Core/SoilAbstractBTreeDataPage.class.st
+++ b/src/Soil-Core/SoilAbstractBTreeDataPage.class.st
@@ -64,15 +64,16 @@ SoilAbstractBTreeDataPage >> insertItem: anItem for: iterator [
 	keyIndex := iterator index hasUniqueKeys 
 		ifTrue: [ self itemIndexForKey: anItem key ] 
 		ifFalse: [ self itemIndexForKey: anItem key value: anItem value ]. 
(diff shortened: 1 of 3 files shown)


MarcusDenker:
Not sure, the problem is that if we now call addItem: in a uniqeKey index, we could end up with multiple keys. Especially for the BTree index pages, it might be better to use a variant that ovrrides

noha (approved):

noha:
I think we should fix the wording. I did a mistake by introducing duplicate keys because it was the new kid on the block But we should have the naming the other way round. So that there is uniqueKeys as special assumption. Then all method selectors and tests will also name it that way round

PR #1091, opened 2026-02-26
Added support for having read-only transactions

noha:


noha:
@MarcusDenker I did the read-only stuff with a flag for now to experiment a bit. I can make that a strategy later

MarcusDenker (approved):

PR #1100, opened 2026-03-22
Added #soilRecursiveHash as general method to calculate a recursive h…

noha:
…ash for all objects. Implemented fixed-,byte- and variable layout for it to work with basic tests.

### DIFF
--- a/src/Soil-Core-Tests/SoilClusterVersionTest.class.st
+++ b/src/Soil-Core-Tests/SoilClusterVersionTest.class.st
@@ -17,7 +17,7 @@ SoilClusterVersionTest >> testPersistentClusterVersionMemory [
 			initializeFromBehavior: Point]);
 		references: ({ 4 . 5 . 6 } collect: #asSoilObjectId).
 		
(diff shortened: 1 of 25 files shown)


MarcusDenker (approved):
# FixedLayout

Wenn keine hasSoilTransientInstVars, kann man per instVarAt: lesen (kann ich spaeter machen)
	
# VariableLayout >> soilRecursiveHash: anObject with: aSet 

-  besser basicSize und dann basicAt: statt at:
- VariableLayouts koennen auch slots haben (siehe e.g. SparseLargeTable), die sind vor dem array, man muss also auch den code haben wie im FixedLayout

MarcusDenker:
For Pharo11, we have 

testSerializationWeakArray
testSerializationWeakLayout

Failing. Reason is that WeakLayout was not a subclass of VariableLayout, the method needs to be copied there (we do that already for serializing, see WeakLayout>>#soilBasicSerialize:with:)

noha:
It is now green. I will merge that PR with another one that introduces the commit strategy so we can have the automatic tracking as an option for a while

PR #1115, opened 2026-04-03
class comments for the basic serializer and materializer

MarcusDenker:
explain the basic serializer and materializer


src/Soil-Serializer/SoilBasicMaterializer.class.st
```diff
@@ -1,3 +1,8 @@
+"
+The basic serializer implements materializing for the subset of simple objects that the SoilBasicSerializer can write.
```
noha: Should be "The basic materializer..."?

noha (changes requested):

PR #1116, opened 2026-04-15
Multikey remove when duplicated value cross page boundaries 

### LINKED ISSUE #1006: Index with duplicate keys: can not remove key with entries going across multiple pages
This tests fails:

```
 testAddingDuplicateKeyAcrossPagesRemove
		"we add so many douplicate keys that we spill over to another page.
                  Then we delete one key-value from the first and last page. Tests that iterator delete can cross to the next page"

	| capacity iterator page |
	capacity := index headerPage itemCapacity.
	index allowDuplicateKeys.
	iterator := index newIterator.
	
	1 to: capacity do: [ :n |
		iterator at: 1 add: n asDummySoilObjectId ].
	page := iterator currentPage.
	1 to: capacity do: [ :n |
		iterator at: 1 add: (capacity + n) asDummySoilObjectId ].
	
	"we really added to a second page"
	self deny: page identicalTo: iterator currentPage.
	self assert: index size equals: capacity*2.
	
	"we should be able to delete from the first page"
	iterator := index newIterator.
	self assertCollection:  (iterator at: 1) includesAll: {1 asDummySoilObjectId}.
	iterator := index newIterator.
	"removing the first. Fails"
	iterator removeKey: 1 value: 1 asDummySoilObjectId.
	self assert: index size equals: capacity*2-1.
```

The failure happens in the `iterator removeKey: 1 value: 1 asDummySoilObjectId`

The reason is that SoilIndexIterator>>#removeKey:value:ifAbsent: uses  findPageFor:, which does not take the value into account. It will return the wrong page and then not find the key->value there:

```
	indexKey := index indexKey: key.
	"We search for the data page for the key to update the value with a removed ID
	and return the prior value just like #at: would do"
	self findPageFor: indexKey.
```

MarcusDenker:
- add a #findPageFor:value: method to the iterator. It will lookup in the index using the key, which for both BTree and SkipList returns the *last* page where a duplicated key is stored. If we do not find the key->value there, we go the the priorPage until we find the item (or read the headerPage).

- Use this in removeKey:value: ifAbsent: 

testAddingDuplicateKeyAcrossPagesRemove is green for both BTree and SkipList

Fixes   #1006

### DIFF
--- a/src/Soil-Core/SoilSkipListIterator.class.st
+++ b/src/Soil-Core/SoilSkipListIterator.class.st
@@ -77,6 +77,17 @@ SoilSkipListIterator >> findPageFor: indexKey [
 	^ currentPage 
 ]
 
+{ #category : #private }
+SoilSkipListIterator >> findPageFor: indexKey value: anObject [
+	| page |
+	page := self findPageFor: indexKey.
+	
+	[(page items includes: indexKey->anObject) or: [page isHeaderPage] ] 
+		whileFalse: [ page := page previousPageIn: index].
+	
+	^currentPage := page
+]
+
 { #category : #accessing }
 SoilSkipListIterator >> index: aSoilSkipList [ 
 	super index: aSoilSkipList.
--- a/src/Soil-Core-Tests/SoilBTreeTest.class.st
+++ b/src/Soil-Core-Tests/SoilBTreeTest.class.st
@@ -255,6 +255,35 @@ SoilBTreeTest >> testAddingDuplicateKeyAcrossPages [
 	self assertCollection: elements hasSameElements: ids
 ]
 
+{ #category : #'tests - duplicate' }
+SoilBTreeTest >> testAddingDuplicateKeyAcrossPagesRemove [
+		"we add so many douplicate keys that we spill over to another page.
+                  Then we delete one key-value from the first and last page. Tests that iterator delete can cross to the next page"
+
+	| capacity iterator page |
+	capacity := index headerPage itemCapacity.
+	index allowDuplicateKeys.
+	iterator := index newIterator.
+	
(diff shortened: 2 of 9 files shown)


src/Soil-Core/SoilSkipListIterator.class.st
```diff
@@ -77,6 +77,17 @@ SoilSkipListIterator >> findPageFor: indexKey [
 	^ currentPage 
 ]
 
+{ #category : #private }
+SoilSkipListIterator >> findPageFor: indexKey value: anObject [
+	| page |
+	page := self findPageFor: indexKey.
+	
+	[(page items includes: indexKey->anObject) or: [page isHeaderPage] ]
```
noha, line 85: If we don't find an item on a page why is it always on previous pages? Is it because we traverse forward until we find a bigger key? That would explain that current page has only the finding key and bigger ones so scanning backwards is feasible?

MarcusDenker: If the item is not there at all, we we return the previous page and the later code will not find the item and execute the failblock.

I was thinking that maybe a better behavior would be to do error handling here, too. We could even return the itemIndex instead of the page (as we get the page via currentPage). 

This would allow us to avoid scanning the items of the page again.

MarcusDenker: > If we don't find an item on a page why is it always on previous pages? Is it because we traverse forward until we find a bigger key? That would explain that current page has only the finding key and bigger ones so scanning backwards is feasible?

yes, if we continue to add keys and they end up in the next page, the index would be updated to return *that* page when we lookup for the page for a key. So we know that pages after have only bigger keys

Issue #1120, opened 2026-04-21
Serializer: CompiledMethod/CompiledBlock referencing LiteralVariables


MarcusDenker:
When serializing, we currently follow all the objects on the literal frame.

This is not good as this means we serialize LiteralVariables (Globals, class vars) including the values.

What we need to do:
- do not store value
- on materialize, do not use the materialized instance but do a re-lookup of the Variable in the class 
- if that fails due to e.g. the class having been changed, use the materialized instance as a fallback

This shows, too, that serializing methods without the class that they are from can only work if we assume that classes do not change.

MarcusDenker:
TODO:

- [x] re-lookup LiteralVariable when materializing CompiledMethod
- [ ] re-lookup LiteralVariable when materializing CompiledBlock
- [ ] Failure handling: when no var found, create Undeclared

PR #1122, opened 2026-04-22
Serializer: CompiledMethod improvementeds

MarcusDenker:
Fix handling of LiteralVariable and the class binding in the last literal

- do not store the value of LiteralVariables On Materialize:

	- re-lookup class by name on materialize
	- re-lookup all Literal Vars by name in the class


MarcusDenker:
First step of https://github.com/ApptiveGrid/Soil/issues/1120, what is missing is to fail back on creating Undeclareds if we can not re-lookup the variable

noha (approved):

PR #1129, opened 2026-05-05
added migration of behavior registry

noha:


MarcusDenker (approved):
needs a backport to v4, too

PR #1131, opened 2026-05-05
Add a commit strategy to transaction

noha:
draft sketch of a commit strategy:
- removed readOnly instVar
- introduced commitStrategy instVar
- added class foe readOnly, manual tracking and automatic tracking as a sketch


src/Soil-Core/SoilPersistentClusterVersion.class.st
```diff
@@ -2,7 +2,7 @@ Class {
 	#name : #SoilPersistentClusterVersion,
 	#superclass : #SoilClusterRecord,
 	#instVars : [
-		'changed'
+		'fingerprint'
```
MarcusDenker, line 5: The 'changed' flag in SoilPersistentClusterVersion being replaced by the fingerprint. This is true even if we configure manual markDirty:
I would have expected that for manual markdirty, we never need to calculate the recursive hash

MarcusDenker (approved):

src/Soil-Core/SoilPersistentClusterVersion.class.st
noha: Manual markDirty: now adds the record directly on recordsToCommit. And the fingerprint was added early because there was no other way doing it. I was thinking about putting the fingerprints in the commit strategy but then it would need another hashed collections to track it and I'm not sure it is worth it. So maybe the best of both worlds is to keep the instVar in the record (I like the idea to move both together around) and only calculate the hash when automatic tracking is on

PR #1133, opened 2026-05-06
a method at:ifAbsentPut: done. 

### LINKED ISSUE #1132: at:ifAbsentPut:
This kind of method (like for ordinary dictionary) will be useful in SoilIndexDictionary.

Daxio0:
#issue:1132
Warning this method return the value at the key if it existed or the new value added


MarcusDenker (changes requested):
can you add a test to SoilIndexedDictionaryTest?

There is already SoilIndexedDictionaryTest>>#testAtIfAbsent, atifAbsentPut is tested for the persistentDictionary in SoilPersistentDictionaryTest>>#testAtIfAbsentPut, to the test would be a combination of the two

noha (approved):

PR #1135, opened 2026-05-06
Iterator: do not check for readVersion in size

MarcusDenker:
Do not check for readVersion in the Iterator when calculating the size (and thus not falling back to slow size).

SoilIndexedDictionaryTest>>#testConcurrentIsEmpty will fail

### DIFF
--- a/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
+++ b/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
@@ -302,6 +302,7 @@ SoilIndexedDictionaryTest >> testConcurrentDo [
 { #category : #'tests - concurrent' }
 SoilIndexedDictionaryTest >> testConcurrentIsEmpty [ 
 	| tx1 tx2 tx3 |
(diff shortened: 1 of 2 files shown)


MarcusDenker:
We start a transaction.

Then a second transaction removes the only entry in the index and commits. 

We are checking the situation for transation where the value was not removed.

isEmpty via iterator -> sees the removedID, restores and it is not empty.
isEmpty via index -> sees the removeID item, not empty (but that could be wrong, it is randomly correct here)

size: goes directly to index size, which says: 0, as the headerpage has the size state of the commited transaction, not the not-commited transaction.

So here a commit "shows through" to an earlier transaction.

noha:
I think this is more or less the same I was saying about index page actuality. We check only for keys and restore them. But index is none of that.
What we need to do is to when we access a page we need to check its lastTransaction and if it is newer restore the old page from the journal. 
So this behavior was always the case because it relies on the time a page is read. So this can be merged.
I open another issue for explaining the things we need to do

noha:
Can we skip the tests for now?

noha:
See #1137

PR #1143, opened 2026-05-08
make SoilPersistentDictionary usable as non-root object.

noha:


MarcusDenker (commented):
Looks good.

I wonder if for speed it would be interesting to have some kind of manual option without the overhead (e.g. to fill up a larger dictionary the first time)

noha:
@MarcusDenker What do you mean with "without the overhead" ?

MarcusDenker:
With beAutomaticWriteTracking, it will calulate the recursive hash over the values, while before it was just doing a manual markDirty on at:put:

But no idea of what the speed impact is

PR #1146, opened 2026-05-11
Made it possible to have indexed dictionary that is not root.

noha:
But then the index creation entry is created last not first.  Added mode to add to head or tail of the journal to make important entries go first.

### DIFF
--- a/src/Soil-Core/SoilIndexedDictionary.class.st
+++ b/src/Soil-Core/SoilIndexedDictionary.class.st
@@ -143,8 +143,8 @@ SoilIndexedDictionary >> isEmpty [
 ]
 
 { #category : #testing }
-SoilIndexedDictionary >> isRegistered [
-	^ index isRegistered 
+SoilIndexedDictionary >> isPresentOnDisk [
+	^ index isPresentOnDisk 
 ]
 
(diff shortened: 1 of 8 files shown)


src/Soil-Core/SoilIndexedDictionary.class.st
```diff
@@ -282,18 +282,21 @@ SoilIndexedDictionary >> size [
 { #category : #serializing }
 SoilIndexedDictionary >> soilBasicSerialize: aSerializer [
 	super soilBasicSerialize: aSerializer.
-	aSerializer registerIndex: self
+	aSerializer registerIndex: self.
+	self isRegistered ifFalse: [
```
MarcusDenker: this is a bit odd: we call #registerIndex, but then check if it is registered?

MarcusDenker (approved):

src/Soil-Core/SoilIndexedDictionary.class.st
noha: It's the naming. #registerIndex: sets the index in the cluster. #isRegistered checks of the index has been created. If change it to reflect that

PR #1148, opened 2026-05-12
added a keysAndValuesDo: in SoilIndexedDictionary and the corresponding tests

Daxio0:
Warning : the key will be an integer, if the original key was a ByteArray asByteArray will be needed, and if it was a String, asByteArray asString

### DIFF
--- a/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
+++ b/src/Soil-Core-Tests/SoilIndexedDictionaryTest.class.st
@@ -328,6 +328,35 @@ SoilIndexedDictionaryTest >> testConcurrentIsEmpty [
 	self assert: tx2 root size equals: 1
 ]
 
+{ #category : #tests }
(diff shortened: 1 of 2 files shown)


noha:
This is not a good PR. Indexes should not be used like a collection. An index has the duty to query a subset of the entries and act on that. I don't want to have the whole collection protocol creeping into the indexes because it is not good. We need a bit of that traversal for writing tests at first

Daxio0:
> This is not a good PR. Indexes should not be used like a collection. An index has the duty to query a subset of the entries and act on that. I don't want to have the whole collection protocol creeping into the indexes because it is not good. We need a bit of that traversal for writing tests at first

Thank you for the explanation, it was because i thought that i needed something like that, but actually i just had to use a SoilPersistentDictionary instead to do what i needed

noha:
Fine. We are always tempted to have everything similar and collections are a powerful companion. But often we forget about the costs of something. And in order to make indexes effective they need to enable loading just what is needed. That is why they are there.

PR #1149, opened 2026-05-12
if read-only is in check mode we need to collect with fingerprint.

noha:
Added faster hash for SoilObjectId


MarcusDenker (approved):
ok!

PR #1150, opened 2026-05-13
Added superclass SoilCommitError as there can be multiple errors duri…

noha:
…ng the commit of a transaction and it should be possible to catch any commit error


MarcusDenker (commented):
ok

PR #1157, opened 2026-06-04
Remove class selector from SoilObjectProxy as it is called too often …

noha:
…resolving too many objects. Furthermore it breaks the inspector. We remove it now and should provide a better approach later


MarcusDenker (approved):
I will check the test

PR #1159, opened 2026-06-17
Fix Windows write support (issue #1005)

### LINKED ISSUE #1005: Windows Compatibility Issue: SoilLockableStream Multiple File Open Problem
## Summary

The Soil database's `backup` and `recover` operations fail on Windows due to a fundamental issue: **attempting to open the same file path with write mode from multiple streams simultaneously**. This behavior is permitted on macOS/Linux but fails on Windows due to exclusive file locking.
This blocks essential database operations on Windows platform and affects data integrity and reliability.

### 1. SoilLockableStream Initialization Method

**Location**: `SoilLockableStream >> initializePath:` (src\Soil-File\SoilLockableStream.class.st:105-108)

```smalltalk
initializePath: aStringOrFileReference [
    fileStream := aStringOrFileReference asFileReference binaryReadWriteStream.
    lockRegistry := SoilFileLockRegistry forPath: aStringOrFileReference asFileReference
]
```

This implementation **always opens files with write permissions** via `binaryReadWriteStream`.

### 2. binaryReadWriteStream Implementation

**Location**: `FileReference >> binaryReadWriteStream` (src\Soil-File\FileReference.extension.st:4-8)

```smalltalk
FileReference >> binaryReadWriteStream [
    "Answer a buffered binary write stream on the receiver"
    ^ ZnBufferedReadWriteStream on: (filesystem binaryWriteStreamOn: self path)
]
```

The stream is created in write mode via `binaryWriteStreamOn:`.

### A. During Backup (Soil >> backupTo:)

**Entry Point**: `Soil >> backupTo:` (src\Soil-Core\Soil.class.st:96-107)

```smalltalk
backupTo: aStringOrFileReference [
    | backupSoil |
    backupSoil := self class new
        path: aStringOrFileReference asFileReference;
        destroy;
        initializeFilesystem.
    SoilBackupVisitor new
        target: backupSoil;
        backup: self.
    backupSoil close
]
```

**Problem Flow**:

1. **Source DB Already Open**: The original Soil instance (`self`) has all files open in write mode:
   - Journal files
   - Object files
   - Control file
   - Index files

2. **Backup Target DB Also Opens**: `backupSoil initializeFilesystem` creates and opens new files in write mode

3. **Problems During SoilBackupVisitor Execution**:
   - In `visitJournalFragmentFile:` (src\Soil-Core\SoilBackupVisitor.class.st:150-153), it calls `aSoilJournalFragmentFile copyTo:`
   - At this point, **the source journal file is still open in write mode**
   - File copy operations may attempt additional access to the same file

4. **Critical Problem During Index Copy**:
   In `copyIndexAt:segment:` (src\Soil-Core\SoilBackupVisitor.class.st:55-81):

   ```smalltalk
   sourceSegment := soil objectRepository segmentAt: segmentId.
   sourceIndex := sourceSegment indexManager at: indexId ifAbsent: [ ^ self indexNotFound: indexId ].

   targetSegment := target objectRepository segmentAt: segmentId.
   targetIndex := targetSegment indexManager createIndexWithId: indexId class: sourceIndex class.
   ```

   - Both source and target with **the same segmentId** attempt to open files in write mode
   - On Windows, this fails due to exclusive locks

### B. During Recovery (Soil >> open)

**Entry Point**: `Soil >> open` (src\Soil-Core\Soil.class.st:390-393)

```smalltalk
open [
    self basicOpen.
    journal recover
]
```

**Problem Flow**:

1. **Files Opened in basicOpen** (src\Soil-Core\Soil.class.st:110-140):
   - SoilControlFile
   - SoilObjectRepository (all segment files)
   - SoilBehaviorRegistry
   - SoilPersistentDatabaseJournal
   - SoilMetadata

   All files are opened in **write mode**

2. **journal recover Execution**:
   In `SoilDatabaseRecovery >> recover` (src\Soil-Core\SoilDatabaseRecovery.class.st:84-135):

   ```smalltalk
   fragmentFile := journal openFragmentForLSN: lastCheckpoint.
   checkpointEntry := [ SoilJournalEntry readFrom: fragmentFile stream ]
   ```

   - Journal file is **already open in write mode**
   - Attempts to read from the same file again
   - On Windows, this conflicts with the existing write lock

3. **During Transaction Journal Replay**:
   In `readFragmentFile:` (src\Soil-Core\SoilDatabaseRecovery.class.st:25-51):

   ```smalltalk
   transactionJournal commitIn: soil recovery: true
   ```

   - Writing to object files occurs during recovery
   - May conflict with already-open write streams

### C. SoilFileWithLayout Problem

**Location**: `SoilFileWithLayout >> open` (src\Soil-Core\SoilFileWithLayout.class.st:96-106)

```smalltalk
open [
    | create |
    create := self path exists.
    stream := self path binaryReadWriteStream.
    ...
]
```

- Classes inheriting from SoilFileWithLayout (SoilControlFile, SoilMetadata, etc.) have the same issue
- Always opens in write mode, preventing multiple instance access to the same file

## Affected Components

The following classes are affected:

1. **SoilLockableStream** (src\Soil-File\SoilLockableStream.class.st)
   - Line 106 in `initializePath:` - uses binaryReadWriteStream

2. **SoilBinaryFile** (src\Soil-Core\SoilBinaryFile.class.st)
   - Line 96 in `stream` - uses SoilLockableStream by default

3. **SoilFileWithLayout** (src\Soil-Core\SoilFileWithLayout.class.st)
   - Line 99 in `open` - directly calls binaryReadWriteStream

4. **SoilJournalFragmentFile** (src\Soil-Core\SoilJournalFragmentFile.class.st)
   - Line 242 in `open` - uses SoilLockableStream

5. **FileReference extension** (src\Soil-File\FileReference.extension.st)
   - Lines 4-8 in `binaryReadWriteStream` - fixed to write mode

### 1. Add Read-Only Mode Support

Add read-only mode support to `SoilLockableStream`:
- Introduce new method like `initializePath:mode:`
- Support 'read'/'write'/'readWrite' mode specification via `mode` parameter

### 2. Explicit File Access Mode Declaration

Make required access modes explicit for each operation:
- **Backup**: Source should be read-only, target should be write-only
- **Recovery**: Journal reading should be read-only, data writing should be write-only

### 3. Improve FileReference Extension

Distinguish and appropriately use `binaryReadStream` vs `binaryReadWriteStream`

### 4. Separate Stream Instances for Different Purposes

Consider maintaining separate stream instances for:
- Reading existing data
- Writing new data
- Recovery operations

### Unix-like Systems (macOS, Linux)
- Allow multiple file descriptors to the same file with write permissions
- File locking is advisory by default
- Multiple processes/threads can write to the same file simultaneously (though not recommended)

### Windows
- Enforces mandatory file locking
- When a file is opened with write access, it is typically locked exclusively
- Other attempts to open the same file fail with "The process cannot access the file because it is being used by another process"

## Design Note

The Windows issue is revealed by **filesystem-level exclusive control differences**, but there is also room for design improvement. The current implementation opens all file accesses in write mode, causing integriry problems in `backup` and `recover` operations that require multiple accesses to the same file.

A fundamental solution requires the ability to appropriately select file access modes based on the operation being performed. This includes:

1. Using read-only mode for source files during backup
2. Using read-only mode for journal files during recovery
3. Properly coordinating write access when multiple components need to access the same file
4. Implementing platform-specific handling where necessary

## Impact

This issue affects:
- **Backup operations** - Cannot create backups on Windows
- **Recovery operations** - Cannot recover from journal on Windows
- **Database reliability** - Critical operations fail silently or with unclear error messages
- **Cross-platform compatibility** - Code works on Unix-like systems but fails on Windows

riverdusty:
## Why

Soil could not be opened or written on **Windows**. Reproducing on a live Pharo image revealed three layered blockers, each only reachable after fixing the prior one:

1. **`fsync`** — `BinaryFileStream>>fsync` called POSIX `fileno`/`fsync` (absent on Windows, giving `SymbolNotFoundError` on `msvcrt`).
2. **File locking** — `OSPlatform>>flockClass` had no Windows implementation.
3. **Multiple file open (#1005)** — the same path was opened in write mode from a second stream, which Windows rejects with `CannotDeleteFileException`.

Fixes #1005.

## What changed

**Windows prerequisites (1 & 2)** — ported from the unmerged `integrate-windows-locking` branch (those `Soil-File` files were unchanged on `main` since the merge-base, so taken as-is):
- `SoilWindowsFileLock` (`LockFileEx`/`UnlockFileEx` + `FlushFileBuffers`), `SoilWinOverlappedStruct`, `WinPlatform` lock/sync
- `BinaryFileStream>>fsync` now delegates to `OSPlatform>>syncFile:`; `fileno:`/`fsync:`/`syncFile:` moved onto the platform classes
- lock classes renamed `UnixFileLock`/`MacOSFileLock` to `SoilUnixFileLock`/`SoilMacOSFileLock`

**Issue #1005 — read-only opens / single write-owner per path:**
- New read-only stream support: `SoilLockableStream class>>readOnlyPath:` + `FileReference>>binaryReadOnlyStream` (in-image locking only — a read-only handle cannot hold an OS write lock, and these reads are of a consistent snapshot)
- Journal **recovery**, journal **query/replication** (`transactionJournalsStartingAt:do:`), and the read-only **journal visitor** (`SoilBasicVisitor>>visitJournalFragmentFile:`) now open fragment files read-only and close them
- Recovery closes its read-only handle before the journal reopens the file for writing; `cycleFragmentFileFrom:` cycles from a read-only recovery file without flushing/syncing it
- **Backup** reuses the already-open behavior-registry `identifier` index instead of opening the same file a second time
- **Index commit** (`SoilNewSkipListIndexEntry`/`SoilNewBTreeListIndexEntry>>commitIn:`) reuses an already-open index instead of re-`initializeFilesystem`

## Verification

Verified end-to-end on Windows (Pharo image): create -> write -> close -> reopen -> write -> reopen -> read round-trips with data intact. Suites green: `SoilBackupTest` (7/7), `SoilDatabaseJournalTest` incl. replication (9/9), `SoilTransactionTest` (9/9); `testCheckpointEmptyRecordsToCommit` and `testConsistency` pass.

## Notes for reviewers

- Non-Windows behavior is unchanged: the new code paths only switch read-only callers to read-only opens and reuse already-open handles.
- **Known follow-up (not in this PR):** the full test *suite* on Windows still shows cascading failures from **test-isolation pollution** — a test leaks a file handle and a later test's `destroy` then hits `CannotDeleteFileException`. Every such test passes when run individually. `SoilLockableStreamTest` also has a pre-existing, unrelated Windows path bug.

Generated with [Claude Code](https://claude.com/claude-code)

---

**Update:** added a second commit fixing a related Unix-only gap: `OSPlatform current syncFile:` dispatches to `Unix64Platform` on 64-bit Unix/Linux, which had no `syncFile:` implementation, causing a `doesNotUnderstand` on any commit that triggers an fsync (including first-time index file creation). Fixed by adding `Unix64Platform>>syncFile:` (mirroring the `WinPlatform>>syncFile:` override pattern) backed by a new `LibC>>fsync:` / `LibC class>>fsync:` convenience pair calling POSIX `fsync(2)` directly.

---

**Update: validated end-to-end on real Ubuntu 64-bit hardware**, not just the Windows reproduction above. Loaded this branch into a clean Pharo 13.1 image (alongside a `main`-baseline run for comparison) and ran the full `Soil-Core-Tests`, `Soil-File-Tests`, and `Soil-Serializer-Tests` suites.

This caught a real regression in the second commit: `Unix64Platform>>syncFile:` passed a `FILE*` (`BinaryFileStream>>fileHandle`) directly to `fsync(2)`, which expects an integer fd — failed with `EBADF` on real 64-bit Linux, dropping `Soil-Serializer-Tests` from 100/100 to 78/100. Fixed by removing that override entirely; `Unix64Platform` now inherits the already-correct `UnixPlatform>>syncFile:` (which converts via `fileno(3)` first). Also tidied a stray `#` in `SoilUnixFileLock`/`SoilMacOSFileLock`'s `fcntl:command:struct:` FFI call spec from the original rename (cosmetic, not the cause of anything, but didn't match convention).

After that fix, comparing this branch against `main` head-to-head on the same Linux box:
- `Soil-File-Tests`: 26/26 (16 existing + 10 new `SoilWindowsFileLockTest`, harmless on Linux)
- `Soil-Serializer-Tests`: 100/100, matching `main`
- `Soil-Core-Tests`: **zero new failures** — every failing/erroring test on this branch (10 failures + 1 error, all lint checks or pre-existing flaky/unrelated tests) was already failing identically on `main`. And **8 pre-existing failures on `main` are now fixed** by this PR, including exactly the kind of locking/journal issues it targets: `cannot memory lock ... journal/0000000000` (×3), `unknown journal entry discovered` in replication, an LSN mismatch in recovery, and `testEnabledFsync`. So this PR also fixes latent bugs on the "officially supported" Linux platform, not just Windows.


noha:
@riverdusty Thanks for the PR. I read you migrated that PR from the existing one. That is perfect and it means we could rebase your PR on the branch the existed to see the diffs?

noha:
The PRs have been disected and integrated. Thanks for the contribution. The changes are now in main

PR #1160, opened 2026-07-12
Some trivial cleanups

MarcusDenker:
- categorize uncategorized
- remove SoilGraphImporter>>#newId which accesses an ivar that does not exist (method not used)
- increase defaultTimeLimit for SoilCleanCodeTest


MarcusDenker:
Pharo 14 build fails, I opened an issue: https://github.com/ApptiveGrid/Soil/issues/1161

rest is green, should not hold up merging

noha (approved):

PR #1164, opened 2026-07-16
Fix SoilBehaviorRegistry detectIndex: NotFound crash

noha:
SoilBehaviorRegistry caches a behavior's version history in memory once (loadHistoryForBehaviorWithIndex:) and never refreshes it. If a new behavior version gets committed for a class after that chain was already cached, later lookups against the stale chain can fail with a raw "not found in OrderedCollection" error from detectIndex:, or with SoilObjectNotFound from behaviorDescriptionWithIndex:andVersion:, surfacing as an unhandled 500 in production when reading old/migrated objects.

Make both lookups self-heal: if no match is found in the cached chain, drop the cached entry, reload the history fresh from the segment, and retry once before raising a proper error.

This is a stopgap fix, not the underlying redesign (immutable SoilBehaviorDescription + cache invalidation on commit is still needed and tracked separately).

Un-skip SoilMigrationTest>>testSerializingObjectWithChangedShapeAfterReading, which reproduces this exact failure and now passes.


MarcusDenker (commented):
ok

PR #1167, opened 2026-07-17
Fix/backup full version history

noha:


MarcusDenker (commented):
backup was meged in  https://github.com/ApptiveGrid/Soil/pull/1169

PR #1168, opened 2026-07-20
Fix silent data loss from objectMap mutation during prepareCommit scan

noha:
SoilAutomaticTrackingStrategy>>prepareCommit (and the analogous SoilReadOnlyStrategy>>prepareCommit) iterated transaction objectMap directly via #do: while deciding which records changed. If handling one record causes further objects to become known to the transaction (e.g. a lazily materializing reference), objectMap grows while it is being scanned. Pharo's HashedCollection>>do: does not tolerate that: growing a Dictionary mid-iteration can silently skip entries that were already there, so a changed record can be dropped from the commit without any error - a silent data loss bug found while verifying the Exclusive->MultiChoice-Gateway migration.

Fix: iterate transaction objectMap values (a stable snapshot array) in both prepareCommit methods, immune to any later growth of objectMap.

Added SoilTestGrowingRecord and
SoilTransactionTest>>testPrepareCommitSurvivesObjectMapGrowingDuringScan, which reproduces the corruption directly against objectMap (reliably drops records against the old code, never against the fix). Verified via mcp__pharo__run_tests: SoilTransactionTest, SoilTest, SoilRecursiveHashTest, SoilMultiVersionTest, and the full Soil-Core-Tests package (559/559 green, pre-existing unrelated self-skip tests aside).

### DIFF
--- a/src/Soil-Core-Tests/SoilTestGrowingRecord.class.st
+++ b/src/Soil-Core-Tests/SoilTestGrowingRecord.class.st
@@ -1,0 +1,36 @@
+Class {
+	#name : #SoilTestGrowingRecord,
+	#superclass : #Object,
+	#instVars : [
+		'onHasChanged',
+		'changed'
+	],
+	#category : #'Soil-Core-Tests-TestData'
+}
+
+{ #category : #converting }
+SoilTestGrowingRecord >> asNewClusterVersion [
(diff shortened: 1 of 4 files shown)


MarcusDenker (commented):
ok

PR #1169, opened 2026-07-20
Fix SoilBackupVisitor to copy full version history for all segments

noha:
visitPersistentClusterVersion: only copied the full MVCC version chain for the meta segment (behavior descriptions), and copied just the current version for regular object segments. A backup must be a byte-identical copy of the source database, including history that old readers (readVersion-based snapshot isolation) still depend on - not just the latest version. copyAllClusterVersions: was already segment-agnostic, so the fix removes the special-casing and always copies the full chain.

Added SoilBackupTest>>testBackupPreservesFullVersionHistoryForRegularObjects, which commits two versions of an object, takes a backup, and asserts that a transaction opened against the backup at version 1 still reads the old value. Fails against the previous code (returns nil/broken data), passes against the fix. Full SoilBackupTest suite: 8/8 passing.

### DIFF
--- a/src/Soil-Core-Tests/SoilBackupTest.class.st
+++ b/src/Soil-Core-Tests/SoilBackupTest.class.st
@@ -160,6 +160,39 @@ SoilBackupTest >> testBackupOldVersion [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1177, opened 2026-07-24
Track B PR1: extract copyClusterVersionChain: reusable chain-copy pri…

noha:
…mitive

Extract the version-chain copy loop from SoilBackupVisitor>>copyAllClusterVersions: into a reusable, parameterizable method copyClusterVersionChain:. It copies a full version chain (ordered newest->oldest, as allVersionsAt: returns) into the target and rebuilds previousVersionPosition there, returning the position of the current (newest) version. copyAllClusterVersions: keeps the single-version fast path and now delegates the multi-version case.

Purely additive: makeCopy:/copyCluster:/copySourceCluster: unchanged, no format change, no smallestReadVersion gate, no blocks. Backup behavior is identical.

Tests (SoilBackupTest):
- testCopyClusterVersionChainRoundtrip: a 3-version chain copied into a fresh target answers newTransactionForVersion: identically to the source for every reader version.
- testCopyClusterVersionChainRebuildsPreviousVersionPosition: the copied chain in the target is backward-reachable from the current version and terminates at previousVersionPosition 0 (invariant I4).

### DIFF
--- a/src/Soil-Core-Tests/SoilBackupTest.class.st
+++ b/src/Soil-Core-Tests/SoilBackupTest.class.st
@@ -304,7 +304,65 @@ SoilBackupTest >> testBackupWithIndexRemoval [
 ]
 
 { #category : #tests }
-SoilBackupTest >> testSimpleBackup [ 
+SoilBackupTest >> testCopyClusterVersionChainRebuildsPreviousVersionPosition [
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1179, opened 2026-07-24
Track B PR2: add smallestReadVersion gate to the version-chain copy

noha:
Add SoilObjectRepository>>copyVersionChain:upToReadVersion: - a gated variant of
copyVersionChain: that drops version-chain entries no reader can still observe.

A version is discardable once the smallest read version has moved to or past the db version that superseded it, or when there is no open transaction at all. This mirrors the existing MVCC cleanup rule in SoilIndexCleaner>>canDiscardOldState: (readVersion isNil or: [readVersion >= supersededAt]). The current (newest) version is always kept; nil (no open transactions) keeps only it. The kept suffix is delegated to copyVersionChain:, so the previousVersionPosition rebuild (I4, oldest kept -> 0) is reused unchanged.

Purely additive: the ungated copyVersionChain: (used by backup for full history) is untouched. No production caller yet - the gate is consumed by tests here and will be wired to soil transactionManager smallestReadVersion in the first real Track-B GC PR.

Tests (SoilObjectRepositoryTest, on a 3-version chain at db versions 1/2/3):
- readVersion 2 keeps {v3,v2}, drops v1, oldest kept previousVersionPosition = 0.
- readVersion 1 keeps all three (a reader at 1 still needs v1).
- nil keeps only the current version.

### DIFF
--- a/src/Soil-Core-Tests/SoilObjectRepositoryTest.class.st
+++ b/src/Soil-Core-Tests/SoilObjectRepositoryTest.class.st
@@ -7,6 +7,12 @@ Class {
 	#category : #'Soil-Core-Tests-Model'
 }
 
+{ #category : #helpers }
+SoilObjectRepositoryTest >> keptVersionsIn: aSoil at: objectId [
+	^ ((aSoil objectRepository segmentAt: objectId segment) allVersionsAt: objectId index)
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1182, opened 2026-07-26
Track A PR2b-1: extract SoilCopyingVisitor common superclass (backup …

noha:
…+ vacuum)

Backup and the coming GC vacuum are two applications of the same "traverse from root and copy each reachable object's version chain into a target". Extract that shared core into a new common superclass SoilCopyingVisitor (SoilVisitor -> SoilCopyingVisitor -> SoilBackupVisitor), so the vacuum can be a sibling rather than a backup subclass.

Moved up from SoilBackupVisitor (unchanged behaviour):
- copyAllClusterVersions: / copyClusterVersionChain: - fetch the chain and, over the kept versions oldest-to-newest, write them to the target segment and follow their references + indexes.
- visitPersistentClusterVersion: - copy every visited cluster's chain.

The differences between applications become seams:
- keptVersionsOf: - which versions to copy. Default (superclass) = the whole chain, which is exactly what backup needs; a pruning subclass (the vacuum) will override.
- targetSegmentFor: - the segment to copy into (understands copyVersionChain: since PR2a). subclassResponsibility; backup returns target objectRepository segmentAt:.
- copyIndexAt:segment: - index persistence. subclassResponsibility; backup keeps its current rebuild-into-target implementation.

SoilBackupVisitor now subclasses SoilCopyingVisitor, drops the three moved methods, and adds targetSegmentFor:. Pure refactor - SoilBackupTest stays 12/12, SoilObjectRepositoryTest 13/13. Depends on PR2a (segment-level copyVersionChain:).

### DIFF
--- a/src/Soil-Core/SoilBackupVisitor.class.st
+++ b/src/Soil-Core/SoilBackupVisitor.class.st
@@ -1,6 +1,6 @@
 Class {
 	#name : #SoilBackupVisitor,
-	#superclass : #SoilVisitor,
+	#superclass : #SoilCopyingVisitor,
 	#instVars : [
 		'target'
 	],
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1183, opened 2026-07-26
Track A PR2b-2: offline object-heap vacuum (Soil>>vacuum)

noha:
The first working GC: rebuild SoilGarbageCollectVisitor as a second SoilCopyingVisitor subclass (sibling of SoilBackupVisitor - backup is untouched) that compacts the object heap in place.

Soil>>vacuum runs it quiescently: capture smallestReadVersion once; for each object segment make a makeGCClone and carry its indexes/ directory over verbatim (indexes map key->objectId, position-independent, so they stay valid after heap compaction); traverse from root copying each reachable object's gated version chain into the clone; then replaceWith: each segment. The whole run is wrapped in ifCurtailed: that deletes the clones, so a failure before the swap leaves the database untouched.

The SoilCopyingVisitor seams do the steering: keptVersionsOf: prunes via segment keptVersionsOf:upToReadVersion: (drop versions no open reader still needs; nil = keep only current), targetSegmentFor: returns the clone, copyIndexAt: is a no-op (indexes carried verbatim; index-value reachability comes from the inherited traversal), and processBehaviorDescription: is a no-op (meta segment left untouched - I2 deferred).

Scope: object heap only. Deferred (unchanged): meta/behavior-description tail-trim (I2), SoilConsistencyVisitor/SoilCompareVersionsVisitor gating (I3/I6), updateLastGarbageCollect, removing the Soil>>garbageCollect prototype guard, multi-segment crash markers.

Tests (SoilGarbageCollectorTest): current version kept + old versions reclaimed (file shrinks); a version an open reader needs survives; an unreachable object is dropped; an indexed SoilIndexedDictionary and its index-only-reachable value still resolve; no clone directory left behind; and an injected mid-run failure (SoilGarbageCollectorFailingVisitor) leaves the database intact. SoilBackupTest stays 12/12.

### DIFF
--- a/src/Soil-Core-Tests/SoilGarbageCollectorFailingVisitor.class.st
+++ b/src/Soil-Core-Tests/SoilGarbageCollectorFailingVisitor.class.st
@@ -1,0 +1,11 @@
+Class {
+	#name : #SoilGarbageCollectorFailingVisitor,
+	#superclass : #SoilGarbageCollectVisitor,
+	#category : #'Soil-Core-Tests'
+}
+
+{ #category : #testing }
+SoilGarbageCollectorFailingVisitor >> keptVersionsOf: aCollectionOfClusterVersions [
+	"test support: fail during the copy phase, after clones are built, before replace"
+	Error signal: 'injected vacuum failure'
+]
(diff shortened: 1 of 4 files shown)


MarcusDenker (commented):
ok

PR #1185, opened 2026-07-26
Track A PR4: opt-in Soil>>checkConsistency (I3/I6), separate from gar…

noha:
…bageCollect

Make the two orphan check visitors usable as an explicit, opt-in structural check - deliberately NOT wired into garbageCollect, which stays lean/fast (a GC run should just GC). Consistency checking is a separate operation you call when you want it (after a GC, in CI, manually).

- SoilConsistencyVisitor>>validate: aSoil - run check over aSoil, raise if errors were found (symmetric with the existing SoilCompareVersionsVisitor>>validate:).
- Soil>>checkConsistency - runs both: index<->heap consistency (I3) + version consistency (I6); raises on the first violation.

A truly preventive gate (check the clone *before* the replace, so a bad clone is never swapped in) is deferred: the check visitors run over a whole Soil, not a bare makeGCClone segment, so that needs a separate clone-level check - checkConsistency will be its building block.

Tests (SoilConsistencyTest): checkConsistency passes on a healthy DB and after a garbageCollect (so the GC provably preserves I3/I6, without slowing the GC itself); raises on an injected version mismatch (I6); and validate: raises on an injected inconsistency via a SoilConsistencyFailingVisitor stub (I3). SoilBackupTest stays 12/12.

Depends on PR3 (garbageCollect entry).

### DIFF
--- a/src/Soil-Core-Tests/SoilConsistencyFailingVisitor.class.st
+++ b/src/Soil-Core-Tests/SoilConsistencyFailingVisitor.class.st
@@ -1,0 +1,11 @@
+Class {
+	#name : #SoilConsistencyFailingVisitor,
+	#superclass : #SoilConsistencyVisitor,
+	#category : #'Soil-Core-Tests'
+}
+
+{ #category : #testing }
+SoilConsistencyFailingVisitor >> checkCluster: cluster [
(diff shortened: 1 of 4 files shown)


MarcusDenker (commented):
ok

PR #1189, opened 2026-07-30
Boolean is missing a recursive hash calculation. Without that it bump…

noha:
…s into the tracked set not recognizing changes


MarcusDenker (commented):
ok

PR #1195, opened 2026-08-22
Fix Soil>>printOn: always printing empty brackets

AndrzejProchyra:
The ifTrue:ifFalse: computed the database version, or #closed, and then discarded the result rather than writing it to the stream. 

Result:
* An open database printed as 'Soil [] /path/to/db' with the version never shown, 
* A closed database printed as 'Soil [] /path/to/db' with the word "closed" never shown

Fix:
Write to aStream in both branches.

### DIFF
--- a/src/Soil-Core-Tests/SoilTest.class.st
+++ b/src/Soil-Core-Tests/SoilTest.class.st
@@ -411,6 +411,36 @@ SoilTest >> testMigrationVersionsHalf [
 ]
 
 { #category : #tests }
+SoilTest >> testPrintString [
+
+	self
+		assert: soil printString
+		equals: 'Soil [0] ' , soil path pathString
+]
+
+{ #category : #tests }
+SoilTest >> testPrintStringAfterCommit [
+
+	soil newTransaction
+		root: Object new;
+		commit.
+
+	self
+		assert: soil printString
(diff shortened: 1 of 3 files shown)


noha:
Thanks for the contribution! I just learned that there ist Stream>>#print: but I find it a bit strange. Is it ok if we do just anObject printOn: aStream instead?

AndrzejProchyra:
> Thanks for the contribution! I just learned that there ist Stream>>#print: but I find it a bit strange. Is it ok if we do just anObject printOn: aStream instead?

Of course 🙂. The reason that I did `print:`, I believe, was that the existing code triggers the Renraku rule `ReNoPrintStringInPrintOnRule`, with the rationale:

> Use the stream argument to #printOn: e.g.( stream print: anObject ) instead of creating an extra stream by using printString. e.g.( stream nextPutAll: anObject printString ).

AndrzejProchyra:
Hmm... Iceberg seems to be reordering the Tonel alphabetically for me, so there is an inadvertent change to `src/Soil-Core/SoilReadOnlyStrategy.class.st`. Is this reordering OK to include here?

noha:
Yes, very annoying that methods are reordered. But not a problem because there is only one

PR #1197, opened 2026-09-02
Finish the application migration so it can be run on a real database

noha:
migrateUpToVersion: started its loop at persistentVersion, which is the version that has already been applied, so the first step was executed a second time. availableVersions meanwhile filtered on > persistentVersion, so the two disagreed with each other. Start the loop after the persisted version, and initialize a fresh control file with application version 0 so version 1 stays a step like any other.

migrate asked for lastVersion, which failed with a SubscriptOutOfBounds on an up to date database because availableVersions is empty then. lastVersion is replaced by targetVersion, which falls back to the persisted version, and migrate returns early unless needsMigration finds something to do. Both targetVersion and the existing persistentVersion are meant to be overridden by applications that determine their current or target version differently, e.g. from the code base.

testMigrationFull/Half/HalfForced are unchanged and still green. testMigrationVersionsFull/Half now expect the leading version 1, since a fresh database no longer counts it as applied. SoilBackupTest green.

### DIFF
--- a/src/Soil-Core-Tests/SoilTest.class.st
+++ b/src/Soil-Core-Tests/SoilTest.class.st
@@ -388,7 +388,7 @@ SoilTest >> testMigrationHalfForced [
 { #category : #tests }
 SoilTest >> testMigrationVersionsFull [
 	soil fabric applicationMigrationClass: SoilTestApplicationMigrationFull.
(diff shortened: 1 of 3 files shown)


MarcusDenker (commented):
ok

PR #1198, opened 2026-09-02
Let applications supply the transaction a version step runs in

noha:
The migration created and committed a transaction per version step itself. An application that already holds an open transaction had no way to run its steps in that one. #newTransaction, #commitTransaction: and #abortTransaction: are now the single places where this is decided, so such an application overrides them to hand in its transaction and leave the commit to its own caller. The default is unchanged: a transaction per step, committed as soon as the step succeeded.

#didMigrateVersion:transaction: is called once a step is applied and the persistent version has been advanced, and updates the metadata's last modified timestamp, which the migration did not do before.

Soil>>migrateApplication runs the migration without the caller having to build one, next to the existing migrateDatabaseFormat.

Records on #versionPragmas that pragma lookup does not follow the superclass chain, so version steps have to live in the class that is instantiated.

SoilTest>>testMigration{Full,Half,HalfForced,VersionsFull,VersionsHalf} green. Verified that a subclass handing in one transaction for all steps persists them together once its caller commits.

### DIFF
--- a/src/Soil-Core-Tests/SoilTest.class.st
+++ b/src/Soil-Core-Tests/SoilTest.class.st
@@ -386,6 +386,21 @@ SoilTest >> testMigrationHalfForced [
 ]
 
 { #category : #tests }
+SoilTest >> testMigrationKeepsVersionsBeforeFailingOne [
+	| txn |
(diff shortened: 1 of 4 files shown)


MarcusDenker (commented):
ok

PR #1199, opened 2026-09-02
Report what the application migration is doing

noha:
The migration ran silently, so an application that wanted to see the progress of a database upgrade had to build its own reporting around it. It now reports through #log:, which emits a SoilSignal, and applications that collect their log elsewhere override that single method.

The signal carries the database it is talking about, so the lines of several databases upgraded in one run stay distinguishable. It is taken from Soil>>id and named databaseId after it.

Reported are: that a migration is needed and up to which version, or that there is nothing to do; every version that is about to be applied; and that migrateUpToVersion: stops before a version which is not applied automatically.

#log: wraps the string with SoilSignal>>on: rather than passing it to #emit:, which leaves the signal without a target and prints a trailing nil after every line.

### DIFF
--- a/src/Soil-Core/Soil.class.st
+++ b/src/Soil-Core/Soil.class.st
@@ -369,6 +369,13 @@ Soil >> migrateDatabaseFormat [
 		migrate
 ]
 
+{ #category : #accessing }
+Soil >> name [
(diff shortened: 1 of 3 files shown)


MarcusDenker (commented):
ok

PR #1202, opened 2026-09-02
Separate the version to migrate to from the versions that may run now

noha:
#targetVersion answered the highest of #availableVersions, and that collection is a list of candidates: versions above the persistent one, cut off before the first that is not applied automatically. Taking the target from it made a version that has to be applied by hand indistinguishable from being finished. Migrating the half automatic test data stopped after version 2 and then reported "application version already at 2, nothing to migrate", although versions 3 and 4 were waiting and the stop was never even logged, because the loop never reached the version it should have stopped at.

The target is now its own decision, taken on the class. A version method is written and tested first and armed afterwards, through #targetVersion:, which makes the live version a deployment artefact rather than a consequence of the code being loaded. It is held in a class instance variable, so arming one migration does not arm another.

Unarmed, #targetVersion answers #defaultTargetVersion, which defaults to #maxVersion: the highest version the version methods declare, regardless of the auto flag. An application that declares elsewhere which version goes to production overrides #defaultTargetVersion with that number and keeps #targetVersion: working, which it would not if it overrode #targetVersion itself. An instance can still be given a target of its own for a single run.

The auto flag only decides whether the target is reached, which is where it was always evaluated. The half automatic run now reports "needs to be migrated to 4", stops at 3 and says so, keeps answering #needsMigration, and reports the stop again the next time.

#maxVersion reads no database and answers zero when a class declares no version at all, so it can be asked before there is a database — class side included, for a caller that has no migration at hand and wants to stamp a freshly created database with the version its code is at.

#availableVersions tests the auto flag only for versions above the persistent one. A non-auto version that has been applied long ago used to set the stop flag and hide everything behind it; asking about versions that are applied already answers nothing about what may run now.

The "nothing to migrate" line reports the persistent version rather than the target, which is what it is talking about.

Four tests: that an applied non-auto version does not block the versions behind it, that a run stops at a non-auto version without the database looking migrated, that a target given to an instance is honoured, and that arming a class stops the run there without arming its sibling. The six existing SoilTest>>testMigration* are unchanged and green.

### DIFF
--- a/src/Soil-Core-Tests/SoilTest.class.st
+++ b/src/Soil-Core-Tests/SoilTest.class.st
@@ -401,6 +401,58 @@ SoilTest >> testMigrationKeepsVersionsBeforeFailingOne [
 ]
 
 { #category : #tests }
+SoilTest >> testMigrationStopsAtNonAutoVersionAndKeepsNeedingIt [
+	soil fabric applicationMigrationClass: SoilTestApplicationMigrationHalf.
+	soil newApplicationMigration migrate.
+	"the run stops before the version that is not applied automatically. The
+	database must not look fully migrated afterwards, the target is still the
+	highest version there is"
+	self assert: soil control applicationVersion equals: 2.
+	self assert: soil newApplicationMigration needsMigration.
+	self assert: soil newApplicationMigration targetVersion equals: 4
+]
+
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1207, opened 2026-09-04
Drop the transient declaration for the journal's removed transaction …

noha:
…variable

SoilTransactionJournal declares #transaction as transient although the class carries #index and #entries only. Two clean code tests keep both shapes of a broken declaration from coming back: a declaration written under a name Soil never sends, and a declared name that is no instance variable of the class or of one of its subclasses.


MarcusDenker (commented):
ok

PR #1210, opened 2026-09-04
Fix and unskip SoilIndexedDictionaryTest>>#testAddRandom

### LINKED ISSUE #1206: SoilIndexedDictionaryTest>>#testAddRandom is skipped
SoilIndexedDictionaryTest>>#testAddRandom is failing now.

i

MarcusDenker:
The test was using #asDummySoilObjectId (as it started as a copy of the low level tests).

Rewrite the test to use a nornal object, as we should not care about ObjectIDs at the level of the IndexDictionary. 

fixes #1206


MarcusDenker:
after this is merged, we use asDummySoilObjectId just in the low level tests of BTree, SkipList and IndexIterator

noha (approved):

PR #1212, opened 2026-09-04
Open journal fragment files read back from disk read-only

noha:
A fragment file that is read back from disk is never written to, so it does not need a lockable stream: #openReadOnly opens a plain read stream, whose file handle the operating system opens non-writable. Any write attempt now fails at the stream level. Read-write is reserved for the fragment file currently being written.

The read paths follow it. Recovery reads fragments through #openFragmentFileNumber:, the visitor and #transactionJournalsStartingAt:do: open through #openReadOnly, and fragment files listed from disk open read-only through their lazy #stream - a fragment file reaching that lazy open has never been opened explicitly, and the fragment files written to are opened through #openFragmentFile: without exception.

Recovery used to declare a fragment it had only read the current fragment file, just to reach its file number when cycling to a fresh one. #cycleFragmentFileAfter: does that directly and closes the file without flushing or syncing it, which leaves #currentFragmentFile: unused.

ZnBufferedReadStream>>isOpen completes the protocol SoilBinaryFile expects of a stream.

### DIFF
--- a/src/Soil-Core-Tests/SoilDatabaseJournalTest.class.st
+++ b/src/Soil-Core-Tests/SoilDatabaseJournalTest.class.st
@@ -90,6 +90,46 @@ SoilDatabaseJournalTest >> tearDown [
 ]
 
 { #category : #tests }
+SoilDatabaseJournalTest >> testCurrentFragmentFileStaysWritable [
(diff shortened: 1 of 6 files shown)


MarcusDenker (commented):
ok

PR #1213, opened 2026-09-05
Cover the embedder's database cycle for residents

noha:
SoilResidentTest proves the generic base "without any embedder's registry of open databases": every one of its tests lives inside a single freshly created database. Two contracts only become visible across a close and a reopen, and neither was exercised.

A resident is not what survives. The registry and the workspace are persistent, the SoilResident is not; on reopening, the description mints a fresh instance, and earlier work has to be found in the workspace rather than in the instance.

And SoilResidentDescription>>initializeFrom: is documented as running both when a resident is added to a new database and "from the embedder's migration path when I am added to a database that already exists". Only the first half could be reached before, because the fixture's database was always brand new.

SoilResidentDatabaseCycleTest adds five tests for that: work surviving a reopen in the workspace, every registered resident coming back from the persisted registry, a resident added to an already existing database arriving with its objects created, that addition leaving the residents already there untouched, and no resident process outliving the stop handshake.

It subclasses SoilResidentTest for the fixture and helpers, with its own path so the two cannot meet on disk. The superclass is concrete, so its own tests are not inherited - intended here, and stated in the class comment.

### DIFF
--- a/src/Soil-Resident-Tests/SoilResidentDatabaseCycleTest.class.st
+++ b/src/Soil-Resident-Tests/SoilResidentDatabaseCycleTest.class.st
@@ -1,0 +1,170 @@
+"
+The embedder's cycle, which SoilResidentTest deliberately leaves out: it proves
+the generic base ""without any embedder's registry of open databases"", so every
+one of its tests lives inside a single freshly created database.
+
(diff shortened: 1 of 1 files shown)


MarcusDenker (commented):
ok

PR #1220, opened 2026-09-06
Run the test suite on Windows too

noha:
The Windows file handling work is being split into reviewable pieces, and without a Windows job none of those pieces can be checked where they matter. Ubuntu tells us nothing about a second open of a path that is already open.

Windows is not green yet - six tests fail today - so it runs with allow-failure, the same way Pharo64-14 already does: the signal is there, the build stays honest about Linux. The include entry says to remove it once Windows passes, and that removal is what turns Windows into a supported platform.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,12 +18,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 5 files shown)


MarcusDenker (commented):
ok

PR #1221, opened 2026-09-06
Give the file lock classes the Soil prefix

noha:
MacOSFileLock and UnixFileLock were the only classes in Soil-File without the prefix, and the Windows work adds a third sibling next to them, so the inconsistency would have grown rather than stayed put.

Rename only. Three references follow: MacOSPlatform>>flockClass, UnixPlatform>>flockClass, and the ignore list in SoilTest class>>classNamesNotUnderTest.


MarcusDenker (commented):
ok

PR #1224, opened 2026-09-06
Close the garbage collector's clones before deleting them

noha:
The collect builds a clone of every object segment and, if the run is curtailed, deletes the clones again. It deleted them without closing them first.

POSIX hides this: unlinking an open file works, the directory disappears, and only the handles leak. Windows does not - the delete fails and the clone survives the abort, which is where the CI run noticed it.

A clone that failed halfway may not close cleanly. That must not stop the deletion, and raising there would replace the error being unwound from, so the close is guarded and the delete happens either way.

The new test asserts the clones' object and index files are closed after a curtailed run - it fails without this change, which the existing testGarbageCollectCurtailedLeavesDatabaseIntact does not, because on POSIX the directory is gone whether or not anything was closed.

### DIFF
--- a/src/Soil-Core-Tests/SoilGarbageCollectorTest.class.st
+++ b/src/Soil-Core-Tests/SoilGarbageCollectorTest.class.st
@@ -23,6 +23,25 @@ SoilGarbageCollectorTest >> tearDown [
 ]
 
 { #category : #tests }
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1228, opened 2026-09-06
Close every part, and say so when one of them would not

noha:
A close that stops at its first failure leaves everything behind it open. That is how one file that will not close becomes a handle leak - and on Windows, the reason the next open fails. Soil>>close closed five things in a row unguarded: a failing object repository left the control file, the journal and the metadata open, with nothing said. Reproduced, and now covered.

Collection>>soilCloseAll closes every element, skips nils, and signals the first error after all of them have been tried. Signalling rather than swallowing is deliberate: the caller has to learn the database did not close cleanly. The first error is the one kept, because it is the one that says why.

Applied where a close releases more than one thing: Soil, SoilObjectRepository, SoilObjectSegment and SoilIndexManager. The index manager empties its register inside the semaphore before closing, so a failing index is gone from it either way instead of leaving the rest listed as open.

SoilBehaviorRegistry>>close was the last unguarded one of the twelve: #index is nil until the registry is opened, and stays nil if opening failed, so closing it answered a doesNotUnderstand - in a tearDown that is all anyone ever sees.

Checked and left alone: SoilIndex>>close nils its store and #store rebuilds it lazily, so closing twice makes a throwaway store object rather than failing. No file is opened by it.

### DIFF
--- a/src/Soil-Core-Tests/SoilTest.class.st
+++ b/src/Soil-Core-Tests/SoilTest.class.st
@@ -194,6 +194,21 @@ SoilTest >> testCyclicProblemPharoSet [
 ]
 
 { #category : #tests }
+SoilTest >> testCloseClosesEveryPartEvenAfterOneFails [
+	"a close that stops at its first failure leaves everything behind it open - which
+	is how one file that will not close becomes a handle leak, and on Windows the
+	reason the next open fails"
+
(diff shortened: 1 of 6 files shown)


MarcusDenker (commented):
ok

PR #1230, opened 2026-09-07
Wait out the coalescing window against a deadline

noha:
A collecting window that is waited out with a single wait on the work signal ends as soon as that signal is raised - and the signal is raised by the very work the window exists to collect. Two cases, one cause:

- a signal still standing on entry. SoilSignalRunStrategy>>awaitDue returns at once whenever work is already pending and leaves the signal up, so the following waitForWorkUpTo: returns immediately and the window is skipped entirely. Measured with a 5s window: written after 21ms.
- a change arriving while the window runs. Measured with the same window and a second change after 1000ms: written after 1031ms.

Under sustained load - the hot path the strategy exists for - both degenerate to one step per change instead of one per burst.

awaitDue now waits against a deadline, checking the stop flag every turn and waiting out the remainder otherwise. Only a stop ends the window early, which it still does: with an hour of window left the stop comes back in 57ms.

The existing burst test could not see any of this because it asserts how many steps a burst causes, not when: a collapsed window still produces exactly one step. The new test asserts the time instead, and fails without this change.

isInWindow exposes the deadline, because with the window living in the strategy "has it reached its window" can no longer be read off the resident - its isWorking covers the step and deliberately not the wait before it.

### DIFF
--- a/src/Soil-Resident-Tests/SoilActiveResidentTest.class.st
+++ b/src/Soil-Resident-Tests/SoilActiveResidentTest.class.st
@@ -82,6 +82,46 @@ SoilActiveResidentTest >> testACoalescingWindowTurnsABurstIntoOneStep [
 ]
 
 { #category : #tests }
+SoilActiveResidentTest >> testACoalescingWindowIsNotCutShortByTheWorkItCollects [
+	"the window has to be waited out both when work is already pending as we enter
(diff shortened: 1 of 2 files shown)


MarcusDenker (commented):
ok

PR #1231, opened 2026-09-08
Let a resident say what it is doing, not only when it fails

noha:
Every self log: in the supervisor sat in an error handler, so the normal path said nothing at all. Silence therefore meant either "all is well" or "nothing is running", and the two looked exactly alike. That is the defect - not too little detail, but a purely negative signal.

Soil>>emit: is new and is the one seam per database. Soil-Core had none: the open path emitted for itself, the migration and the supervisor each had their own log:, so an embedder collecting its log elsewhere had to override three places. Now everything reports through the database, and the supervisor and the residents report through it in turn. Built with #on: rather than SoilSignal's #emit:, the way the migration already builds its own - #emit: fills #message and leaves the wrapped object empty, which puts a trailing "nil" on every line.

What is reported: a resident's state changes (started, stopping, stopped, failed to start), a revived process with its count, a failed step with the number of the step it was, a tick summary of the work since the last one, and - the only level whose volume scales with the work - every single step. That last one is on by default and the embedder can turn it down through the supervisor, because Beacon filters by signal class and these lines share one.

A tick with no work says nothing, so silence on a tick means "nothing to do", which is a statement, unlike the silence this replaces.

Every line names its resident. The supervisor's own per-resident lines now go through the resident for exactly that reason, which also removes the duplicate report of a failed start: the state machine says it, the supervisor no longer says it again in a different shape.

SoilTestActiveResident loses its own stepCount and inherits the real one - it was counting steps because the base did not.

### DIFF
--- a/src/Soil-Core/Soil.class.st
+++ b/src/Soil-Core/Soil.class.st
@@ -131,7 +131,7 @@ Soil >> backupTo: aStringOrFileReference [
 { #category : #'opening/closing' }
 Soil >> basicOpen [
 	self isOpen ifTrue: [ self error: 'Database already open' ].
-	('open soil database at ', path asString) soilEmit.
+	self emit: 'open database at ', path asString.
 	control := SoilControlFile new
 		soil: self;
 		open.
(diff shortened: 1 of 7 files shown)


MarcusDenker (commented):
ok

PR #1232, opened 2026-09-08
Say which way recovery went, and what the database did around it

noha:
#recover has six ways out and took all of them in silence. A database that had quietly started a fresh fragment file, one that had replayed a hundred transactions, and one that had nothing to do all looked the same from outside - which is how a day of hunting a Windows problem could pass without the log ever naming which had happened. Three of the six can be staged, and each now says what it did, the bogus-checkpoint one carrying the read error that is the only thing describing what was wrong with the file. A replay counts what it replayed: a line saying only that recovery ran leaves open whether it found one transaction or a thousand.

Around it: #close says so before it starts, because the interesting case is the one where closing does not return; #backupTo: brackets itself; and a new segment says so, which happens once in a database's life.

Checkpoint is deliberately silent, and that is a correction to what the plan assumed. It looks like a rare lifecycle step from its callers in recovery and on a first open, but SoilTransactionManager>>commitTransaction: checkpoints on every commit that changed something. Measured, not read: three commits produced three lines before it was taken out again. A test holds it there.

SoilRecoveryReportingTest is the first test to exercise recovery at all. Staging a crash means winding the control file's checkpoint back, the same handle SoilJournalConverter reaches for, because a clean close never leaves anything past the last checkpoint.

The one resident test that assumed every collected line came from a resident is narrowed: the database reports about itself now too.

### DIFF
--- a/src/Soil-Core-Tests/SoilRecoveryReportingTest.class.st
+++ b/src/Soil-Core-Tests/SoilRecoveryReportingTest.class.st
@@ -1,0 +1,116 @@
+"
+What recovery says about the way it went.
+
+#recover has six ways out and used to take all of them in silence, which is how a
+day of hunting a Windows problem could go by without the log ever naming which
+one had happened. These tests stage the three that can be staged and hold each to
+its own line.
+
+Staging them means putting a database into the state a crash would leave. Every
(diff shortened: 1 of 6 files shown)


MarcusDenker (commented):
ok

PR #1234, opened 2026-09-09
Print the database path, not its inspector name

noha:
'open database at File @ /data/apptivegrid/66/95/2e/...' - the 'File @ ' is FileReference>>printOn:, which #asString goes through. Inherited from the line this replaced, and invisible until it showed up in a production log.


MarcusDenker:
ok

PR #1236, opened 2026-09-09
Report a stop only when there was one

noha:
requestStop put an already stopped resident back into #stopping and out again, and said both. That is a state cycle that did not happen, reported twice, and two stops in a row is not exotic - an embedder that stops explicitly and then closes does exactly that, which is how this turned up in a real log.

markStopped is guarded for the same reason: terminateProcesses marks a resident stopped as well and may run after the loop already did.


MarcusDenker (commented):
ok

PR #1246, opened 2026-09-11
Rename SoilObjectIndexTest to SoilObjectIndexFileTest

noha:
The test class exercises SoilObjectIndexFile exclusively - both test methods instantiate it and assert on its header, prefix and object index. The old name claimed a different subject and kept the name SoilObjectIndexTest occupied for a class it does not test.

No senders exist anywhere in the repository, so the rename is confined to the class definition and its three method definitions.


MarcusDenker (commented):
ok

PR #1248, opened 2026-09-11
fix SoilPersistentDictionary to use TypeCode (and add CleanCodeTests to Serializer)

### LINKED ISSUE #1247: Serializer: SoilSerializer>>#nextPutPersistentDictionary: never called
It seems we never write SoilPersistentDictionary using the TypeCode

MarcusDenker:
- add SoilSerializerCleanCodeTest (copy of the one from Soil)
- fixe some trivialities
- fix SoilPersistentDictinary to use TypeCode

fixes #1247


MarcusDenker:
This is backward comaptible: existing serialized dicts are read as normal objects, new ones are stored with typecode

noha (approved):

PR #1249, opened 2026-09-14
Move an entry to a new key without materializing its value

noha:
A queue changes the due time of an entry by moving it from one key to another, and it does that repeatedly while a backoff runs. Composing that out of #removeKey: and #at:put: reads the referenced object from disk every time: #removeKey: answers an unresolved SoilObjectProxy, #at:put: sends #makeRoot:, and that resolves the proxy through #theNonSoilProxy. Nothing about the key change needs the object.

#moveKey:to: carries the value as its SoilObjectId instead, so the entry moves and the object stays where it is. Both halves run in the caller's transaction, so an abort in between leaves the old key in place.

Moving needs a unique index, otherwise it is undefined which of the entries under the old key travels. A dictionary is unique by definition, so there is nothing to check; the comment says so for whoever lifts this to an index that allows duplicates.


MarcusDenker (commented):
ok

PR #1250, opened 2026-09-14
Indexed dictionary remove durability

noha:


src/Soil-Core/SoilIndexedDictionary.class.st
```diff
@@ -134,6 +134,19 @@ SoilIndexedDictionary >> index [
 
 { #category : #accessing }
 SoilIndexedDictionary >> indexUpdated [ 
+	"This marker is idle and is kept anyway. Nothing reads #dirty but the guard below, and
```
MarcusDenker, line 137: So can we remove it?

MarcusDenker (commented):
ok

PR #1251, opened 2026-09-14
Store a position on an index as a value that outlives its transaction

noha:
A SoilCursor is the key that was read last on a held index, immutable and persistent. An iterator is bound to its transaction, and one kept across commits would hold a read transaction open, which under MVCC stops old versions from being cleaned up. The cursor is what that iterator's current key distills to, and #asIterator wins the iterator back at the same place inside whatever transaction the caller opened.

Reading starts after the cursor's key, excluding it, so the entry that was read last is not handed out twice. A key that is no longer there is still a position: the cursor sets it on the iterator and reading goes on from whatever the iterator makes of a key it cannot find.

#asIterator positions and nothing else - reading stays with the iterator. That keeps the answer of #nextKeyCloseTo: out of reach, which is not nil but the last key it saw when it finds nothing after the key; taken for a position it would silently move the cursor to the end. Positions come from associations that were actually read, via #advancedTo:, which answers a new cursor rather than moving this one - what moves is the holder.

SoilBasicCursor carries the key alone. The second target, a raw index addressed by segment and index id outside a transaction, gets its own subclass when its case turns up: it cannot answer #newIterator under the same contract as a holder that knows its transaction.


MarcusDenker (commented):
ok

PR #1252, opened 2026-09-14
Bound a scan to a section of an index marked off by a key prefix

noha:
Several independent structures can share one index when every key carries a marker in front. What was missing was the piece that says where such a section begins and, above all, where it ends: an index has no idea that it is shared, so a scan reads straight on into the next section unless something stops it.

SoilIndexRange is that abstraction and it is for reading only. Inserting, removing and moving a key stay on the iterator, so the writing side lives in one place rather than two. Nothing is enforced either: the range supplies the stop condition and the reader carries it. A loop that forgets to ask is a programming error a test catches, which is cheaper than virtual views over an index that none of the known users need.

SoilPrefixRange holds the encoding, and both directions of it, because implemented apart they drift: out of a marker a section, out of a key its marker.

Three things follow from the way an index stores a key, and each of them fails in silence rather than loudly. A key is packed into one number of the index's key size, so a shorter key is padded on the *left* and its marker slides out of the front into marker 0 -- hence every key the range builds fills the whole key size, marker first and zero padding last. The marker takes a fixed width, big endian, so numeric and byte order agree. And the width is a parameter rather than a constant: only the application knows how many sections it will ever need, it has to stand when the index is created, and it may be 0, in which case there is no namespace and the structure pays no byte for it.

The scan test also pins down the trap in the positioning call it builds on: past the last key of the index, #nextKeyCloseTo: answers the last key instead of nil and leaves the iterator sitting on it, so a loop that ends only on nil reads that entry for ever.

SoilKeyRange, the sibling that takes a plain pair of bounds, is named in the design but has no user yet and is not built here.


MarcusDenker (commented):
I do not really understand for what his is needed

noha:
The code itself enables to partition an index into sections by a prefix. This is used by the Soil-Queue to make it possible to share an index for mulitple queues and not have to create an index per queue all the time

PR #1253, opened 2026-09-14
Position the cursor behind the last item when no close key is found

noha:
#nextKeyCloseTo: left currentPageIndex at 0 when the key sits behind everything the page holds: #keyOrClosestAfter: answers nil there and #indexOfKey: nil is 0. For #basicNextAssociation a 0 means "no key on this page yet", so it answered the first item of the page. Reading went on at the beginning of the page instead of after the key, the opposite of what the comment of the method promises.

Put the cursor one behind the last item of the page instead. Reading forward now answers nil on the last page and the first item of the following page otherwise, which is the entry the key asked for. Reading backward keeps answering the last item of the page - one behind it rather than on it, so that item is not skipped.

The answered key stays the last one of the page and is therefore before the key rather than after it. That is odd but pinned by #testNextKeyCloseTo and #testNextCloseTo, so it is left alone.


MarcusDenker (commented):
ok

PR #1254, opened 2026-09-14
Run the Windows CI jobs only on push to main

noha:
Pull requests stay Linux-only, the Windows matrix entries are skipped there and run once the change is merged to main.


MarcusDenker (commented):
ok

PR #1255, opened 2026-09-14
Answer the key that actually comes after at #nextKeyCloseTo:

noha:
#nextKeyCloseTo: promised the first entry at or after key and answered the last key of the page instead whenever no key on that page qualified. #nextCloseTo: then looked that earlier entry up over #at: and handed it out as "the next one after key", so both answered something before key rather than after it.

Two cases hid behind that fallback. Behind the last entry of the index there is no answer, and nil is what says so. In the gap between the last item of a page and the first item of the page after it there is an answer, just not on the page #findPageFor: stops at - #keyOrClosestAfter: only ever looks at a single page. So we walk on to the next page that holds something and answer its first key.

The cursor is left on the entry that was answered, the same place a found key leaves it, and one behind the last item when the answer is nil, so that reading backwards still answers that item instead of skipping it. It stayed at 0 there before, which #basicNextAssociation reads as "no key on this page yet": reading went on at the beginning of the page rather than after the key.

The three tests that pinned the old answer with "this is a bit odd, when looking with larger values we get the last one" expect nil now. #testNextCloseTo of SoilIndexedDictionaryTest carried a trailing space in #'15 ' which made that key a byte longer and therefore larger than #'20', so the row tested the same case as #'100' instead of the one between the two entries it sits in - the space is gone.

Callers in apptive-base have to follow: AGListGlobalEventsOperation>>#readOlderThan: gives up on nil and would lose the page it is asked for when before sits behind the whole log.


MarcusDenker (commented):
ok

PR #1257, opened 2026-09-15
Back up a database whose index has not passed the placeholder

noha:
A database with one commit and nothing but its root cannot be backed up at all: SoilBackupVisitor>>visitObjectSegment: copies the source segment's lastObjectIndex into a freshly created target, and #lastObjectIndex: demands strict growth. But a fresh target does not start empty - SoilObjectRepository>>initializeFilesystem gives its first segment lastObjectIndex 1 and its meta segment 2, as placeholders for the root and the meta description. A source that has not yet overtaken those copies a value that is equal, the setter reads that as an index that failed to grow, and the backup dies with "segment indexes can only grow".

Which of the two segments trips first depends on how far the source has grown, so the same database can copy on one image and fail on another - it looked like a Pharo version difference and is not one.

#updateLastObjectIndex: is the tolerant sibling that already exists for exactly this shape: it returns early when the value has not grown rather than refusing it. Copying an index that equals the placeholder is a faithful copy, and the target keeps the higher placeholder, which is what a fresh database has anyway.

This is the state a user database is in between being created and being used for the first time, so it is reachable in production and not only in tests.


MarcusDenker (commented):
ok

PR #1259, opened 2026-09-15
Create a missing fragment file with its header

noha:
The file the last checkpoint names can be gone, and that is how recovery reaches the journal: it cycles the fragment file when the one named is missing. Writing to a name that is not there opens the file into existence through the write stream, without the header and version #createFragmentFile: writes, and leaves behind a fragment file no reader can make sense of.

#currentFragmentFile now creates the file when it is not there instead of opening it.


MarcusDenker (commented):
ok

PR #1261, opened 2026-09-15
Close the journal fragment before deleting it

noha:
Copying the source journal into a backup first clears the target's own fragment. That fragment is opened lazily, the first time anything touches the target's journal, and Windows refuses to delete a file that is still open; on POSIX the unlink succeeds and the open handle keeps the old inode alive, so this only ever showed on Windows.

It is now closed before the delete and forgotten afterwards, so the next write opens a fresh one.


MarcusDenker (commented):
ok

PR #1262, opened 2026-09-16
Refuse a version chain that leads back into itself

noha:
#allVersionsAt: follows previousVersionPosition until it reaches zero and keeps every version it passes. A chain that returns to a position it has already been at therefore never returns, and the collection grows until the image is gone - which is what a database in the estate did to whichever process tried to back it up, for four months, without an error anywhere.

The walk now remembers where it has been and signals SoilDatabaseIsInconsistent when it arrives there again, naming the object and the position. One damaged object then costs its own database rather than the process reading it.

Only this walker is guarded. #basicAt:version: and #at:version: follow the same links but stop at the first version they can use, so they circle only when asked for a version the cycle does not contain; this one follows the chain to its end whatever it finds.


MarcusDenker (commented):
ok

PR #1263, opened 2026-09-21
Mark the page for writing when a key is removed

noha:
Removing a key writes its tombstone straight into the page's items collection, which leaves #needWrite false - unlike #addItem:, which sets it. #decreaseSize does mark the header page, so the counter reached the disk while the data page did not.

While another transaction holds a read version older than the removal, SoilIndexCleaner refuses to clean that page (#canDiscardOldState:) and nothing else marked it for writing. After a reopen the removed key was back and the size counter was short of the entries actually in the index; the next insert on that page then wrote the resurrected entry out for good.

The new test needs its 2000 entries because the key has to land on a data page of its own - with a handful of them everything sits on the header page, which #decreaseSize marks anyway. That is why #testRemoveKeySurvivesReopen next door stayed green.

Found in a production database whose indexes had drifted by 29, 87 and 142 entries, with 177 removed entries back in place.


MarcusDenker (commented):
ok

PR #1264, opened 2026-09-21
Recompute an index's size

### LINKED ISSUE #1087: Size counter gets out of sync
When adding and removing entries to an index the size counter gets out of tune. In my case it said 300 while there were 460 in the index

noha:
#size answers a counter kept incrementally by #increaseSize and #decreaseSize, and nothing ever recounted it. Where the two drift apart there was no way back to a correct size: #sizeViaIteration only counts, #size: on the skip list's header page did not mark the page for writing, and the B-tree's header page had no setter at all.

#recomputeSize counts what the pages hold and stores that. A repair has to go through #thePersistentInstance: a transaction's index is a copy-on-write wrapper whose pages are copies and whose #addDirtyPage: does nothing, so a size written there would never reach the disk.

Also drops the (size = -1) guards from the skip list's header page. They are the placeholder from the days of backwards-compatible index versions and have been dead since Soil v4 - nothing sets -1, nothing reads it.


MarcusDenker (commented):
ok

